Issue #125, opened 2021-12-27
Image converter


Arlodotexe:
**Is your feature request related to a problem? Please describe.**
Sometimes as a dev, I have an image in one format, and quickly need it in another. The only way to do this is
- Using an online converter
- Open in MsPaint and re-save it.

Both of these options are cumbersome and overkill for such a simple thing.

**Describe the solution you'd like**
Similar to the Image Png / Jpeg compressor tool, I'd like a new tool under the "Graphic" category for converting images between types.

**Describe alternatives you've considered**
- Using an online converter
- Open in MsPaint and re-save it.
- Building my own tool (not everyone should have to)

rlm96:
Hey, I would like to work on this improvement if possible 😁

veler:
Hi,
I think @btiteux was about to start working on it. @btiteux , is that correct? If not, we can let @rlm96 giving it a try :)

PR #139, opened 2022-01-17
Image converter tool

rlm96:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: [125](https://github.com/veler/DevToys/issues/125)

## What is the new behavior?

- Created a new tool under the Graphic section for Image Converter
- Tool can lossless convert between different image formats

## Other information

image
image
image

## Quality check


btiteux:
Hello,
I have some recommendation about the Ui. 

I think I prefer to see the final file under the save all button, for example: I want to convert the fileName.png to fileName.jpg under the "save all" button I would like to see the fileName.jpg instead of the original file name.

And I tried to convert the same file in different format I can't do it, maybe it's linked to my previous comment?

If I want to save the new file in the same folder as the original file the original one is overwritten by the new one.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
```diff
@@ -0,0 +1,168 @@
...
+
+        private string _fileSize = string.Empty;
+
+        internal ImageConverterStrings Strings => LanguageManager.Instance.ImageConverter;
+
+        internal string FileName { get; }
+
+        internal string FilePath { get; }
+
+        internal SoftwareBitmap Bitmap { get; }
```
veler, line 42: following my comment below: you will need to make this property nullable.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
```diff
@@ -0,0 +1,168 @@
...
+        internal event EventHandler? DeleteItemRequested;
+
+        internal ImageConversionWorkItem(StorageFile file)
+        {
+            Arguments.NotNull(file, nameof(file));
+
+            FileName = file.Name;
+            FilePath = file.Path;
+
+            using (IRandomAccessStream stream = Task.Run(async () => await file.OpenAsync(FileAccessMode.Read)).Result)
```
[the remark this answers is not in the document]
veler: You can also use a similar approach to what `ComputePropertiesAsync` does.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
```diff
@@ -0,0 +1,168 @@
...
+            };
+
+            await dialog.ShowAsync();
+        }
+
+        #endregion
+
+        private async Task ComputePropertiesAsync(StorageFile file)
+        {
+            BasicProperties fileProperties = await file.GetBasicPropertiesAsync();
```
veler, line 145: Please add at the beginning of this method `await TaskScheduler.Default`. This will make sure we go off the UI Thread.

It's possible we're missing this in PNG / JPEG Compressor too (oops).

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
```diff
@@ -0,0 +1,168 @@
...
+                    await encoder.FlushAsync();
+                }
+
+                DeleteCommand.Execute(this);
+            }
+        }
+
+        #endregion
+
+        #region ShowErrorMessageCommand
```
veler, line 124: This command doesn't seem to be used:
1. Should we consider using it?
2. If not, please remove it.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
```diff
@@ -0,0 +1,249 @@
+﻿using System;
```
veler, line 1: Please add `#nullable enable`.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
```diff
@@ -0,0 +1,249 @@
...
+using Windows.Storage.Pickers;
+using Windows.Storage.Streams;
+using Windows.UI.Xaml;
+
+namespace DevToys.ViewModels.Tools.ImageConverter
+{
+    [Export(typeof(ImageConverterToolViewModel))]
+    public sealed class ImageConverterToolViewModel : ObservableRecipient, IToolViewModel, IDisposable
+    {
+        private static readonly string[] SupportedFileExtensions = new[] { ".png", ".jpg", ".jpeg", ".jfif", ".pjpeg", ".pjp", ".jxr", ".bmp", ".tiff", ".tif", ".heif", ".gif" };
```
veler, line 27: I see that GIF files are supported. I'm afraid this will confuse the user as most of GIF are animated and that the tool would convert just the first frame of the animation. I'd suggest removing the GIF support for now and wait for users feedback to see if people actually want it.

Additionally, AFAIK, jfif, pjpeg, pjp, jxr and heif aren't the most common used image format in development (but I can be wrong). 
I'd suggest either sticking to PNG, JPG, JPEG and BMP or updating `InvalidSelectedFiles` to indicate all these formats. But to be fair, unless someone convinces me otherwise, I think it makes sense to keep it simple and just support PNG, JPG, JPEG and BMP for now. 
What do you think?

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
```diff
@@ -0,0 +1,249 @@
...
+
+        private bool _isSelectFilesAreaHighlithed;
+        private bool _hasInvalidFilesSelected;
+
+        public Type View { get; } = typeof(ImageConverterToolPage);
+
+        public string ConvertedFormat;
```
veler, line 34: Please make this a property with get and set.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
```diff
@@ -0,0 +1,249 @@
...
+        {
+            ConvertedFormat = "PNG";
+        }
+
+        public void Dispose()
+        {
+            DeleteAllCommand.Execute(null);
+        }
+
+        internal ObservableCollection<ImageConversionWorkItem> ConversionWorkQueue { get; } = new();
```
veler, line 73: Please group this property with the others above.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
```diff
@@ -0,0 +1,249 @@
...
+
+            var folderPicker = new FolderPicker
+            {
+                ViewMode = PickerViewMode.List
+            };
+
+            StorageFolder? selectedFolder = await folderPicker.PickSingleFolderAsync();
+
+            if (selectedFolder is not null)
+            {
```
veler, line 211: If I'm not mistaking, even if this method is a Task, we're still running on the UI Thread (you can verify it with `ThreadHelper.ThrowIfOnUIThread`).
This is fine until line 210. We need to be on the UI Thread in order to call PickSingleFolderAsync.

However, since the loop below can potentially treat hundreds files, I'd heavily suggest to run this part off the UI thread, perhaps showing a progress bar if needed (not sure if it really make sense but we can brainstorm it).

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
```diff
@@ -0,0 +1,249 @@
...
+                    }
+
+                    work.DeleteCommand.Execute(null);
+                }
+            }
+        }
+
+        #endregion
+
+        private void QueueNewConversion(StorageFile file)
```
veler, line 230: Since you're doing the conversion of image when the user clicks on Save, I'd argue that either this method should renamed, because we're technically not queuing a conversion yet, or that the UI should be revamped.
It's a little confusing to me that we should "save" in order to "perform a conversion".

veler (changes requested):
Thanks @rlm96 for proposing this change! It's pretty good overall. @btiteux and I are a little bit confused by the UI and user experience however. I think we need to brainstorm it a little bit.

Thank you so much ! Looking forward to see this PR growing so we can include this tool :D

src/dev/impl/DevToys/Helpers/ImageHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Threading.Tasks;
+using Windows.Foundation;
+using Windows.Graphics.Imaging;
+using Windows.Storage.Streams;
+
+namespace DevToys.Helpers
+{
+    public static class ImageHelper
+    {
+        public static bool IsPngFormat(string format) { return format is "PNG"; }
```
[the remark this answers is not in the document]
rlm96: Removed 👍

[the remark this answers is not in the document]
rlm96: Removed 👍

src/dev/impl/DevToys/Helpers/ImageHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+        public static bool IsJpegFormat(string format) { return format is "JPEG" or "JPEGXR"; }
```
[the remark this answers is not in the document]
rlm96: Used string.Equals instead of ```is```

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
rlm96: 1. Not really, if conversion fails an exception will be raised
2. Removed 😄

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
```diff
@@ -0,0 +1,249 @@
...
+                if (storageItems is null || storageItems.Count == 0)
+                {
+                    return;
+                }
+
+                foreach (IStorageItem storageItem in storageItems)
+                {
+                    if (storageItem is StorageFile file)
+                    {
+                        string fileExtension = Path.GetExtension(file.Name);
```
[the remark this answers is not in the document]
rlm96: Same variable is used in [PngJpgCompressorToolViewModel.cs (line 136)](https://github.com/veler/DevToys/blob/770d7cf8ff67a24ff6e9799c9fae963078e4f071/src/dev/impl/DevToys/ViewModels/Tools/Graphic/PngJpgCompressor/PngJpgCompressorToolViewModel.cs#L136), method is similar. I can remove the variable and use storageItem instead, but just to let you know 😄

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
veler: If an exception is raised while doing the conversion, we should probably catch it and tell the user why it failed. :)

src/dev/impl/DevToys/Helpers/ImageHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+        {
+            var encodingOptions = new BitmapPropertySet();
+
+            if (IsJpegFormat(format))
+            {
+                var booleanBitmapTypedValue = new BitmapTypedValue(1, PropertyType.Single);
+                encodingOptions.Add("ImageQuality", booleanBitmapTypedValue);
+            }
+
+            if (encodingOptions is not null)
```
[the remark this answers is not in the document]
rlm96: Ooops, sorry, I have used Count property instead to check there is any encodingOption(s)

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
rlm96: I have renamed the method to ```QueueNewConversionWorkItem```. Please, let me know if you think another name would  fit better.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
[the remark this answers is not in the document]
veler: Good catch, feel free to do this change in PngJpegCompressor too :)

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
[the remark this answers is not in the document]
rlm96: I've applied that approach, just to make sure this is a fire-and-forget strategy so not frozing UI thread implies that if there is any failure in the code execution using ```.Forget()``` we won´t be able to catch it. On the other hand, that non-catched fail will potentially imply a future failure/exception that will be captured and shown to the user, just to mention that 😄

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
rlm96: I added GIF feature just because I was using all the codecs present in the library, but yes, that makes sense.

From my point of view, maybe someone at some point can find usefull the HEIF/JPEG XR/TIFF options. We can limit using ```SupportedFileExtensions``` variable the extensions allowed (just let me know which extensions should be delisted), but keep those encoding options may be a differentiating factor compared to other tools/online options.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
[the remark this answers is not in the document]
veler: Good to mention that. There's a `ForgetSafely` method you can use. It has a callback for when an exception is raised ;)

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
veler: I see :)
For now, let's try to support:
- PNG
- JPG
- JPEG
And let's see if users are asking for more.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
veler: `QueueNewConversionWorkItem` works, yes. :)

src/dev/impl/DevToys/DevToys.csproj
```diff
@@ -461,6 +469,12 @@
     <AppxManifest Include="Package.appxmanifest">
       <SubType>Designer</SubType>
     </AppxManifest>
+    <PRIResource Include="Strings\zh-Hans\ImageConverter.resw" />
```
veler, line 472: You can remove these lines, since line 478 includes everything (I know, have to change the file manually ... :( )

src/dev/impl/DevToys/Helpers/ImageHelper.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Threading.Tasks;
+using Windows.Foundation;
+using Windows.Graphics.Imaging;
+using Windows.Storage.Streams;
+
+namespace DevToys.Helpers
+{
+    public static class ImageHelper
+    {
+        public static bool IsJpegFormat(string format) { return string.Equals(format, "JPEG") || string.Equals(format, "JPEGXR"); }
```
veler, line 13: Please use `StringComparison.Ordinal` in the arguments of `String.Equals`.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/PngJpgCompressor/PngJpgCompressorToolViewModel.cs
```diff
@@ -133,9 +133,7 @@ private async Task ExecuteSelectFilesAreaDragDropCommandAsync(DragEventArgs? par
                     IStorageItem storageItem = storageItems[i];
                     if (storageItem is StorageFile file)
                     {
-                        string fileExtension = Path.GetExtension(file.Name);
-
-                        if (SupportedFileExtensions.Any(ext => string.Equals(ext, fileExtension, StringComparison.OrdinalIgnoreCase)))
+                        if (SupportedFileExtensions.Any(ext => string.Equals(ext, Path.GetExtension(file.Name), StringComparison.OrdinalIgnoreCase)))
```
veler, line 136: I think what @btiteux was suggesting is to use `file.FileType` instead of `Path.GetExtension`.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
rlm96: I have already added only support for PNG, JPEG and BMP (as you included it in the previous comment) formats

JyeGuru: > maybe someone at some point can find usefull the HEIF/JPEG XR/TIFF options

As someone who works in technical services, I regularly get end users sending me photos as HEIF files, which get held in our mail filter (they are classified under "media files" along with HEVC/MP4 and I can't change this). Our teams have tools they use to convert these to JPG and re-attach to the email, but there's nothing standardized (just various online/portable apps). I would definitely use this feature for these files if it were available.

uahnbu:
The compressor tool doesn't work quite well. Some images aren't even compressible at all. Maybe we should add a compressing percentage slider?

veler:
> The compressor tool doesn't work quite well. Some images aren't even compressible at all. Maybe we should add a compressing percentage slider? And yeah we could show the diffmap between original and compressed image.

I feel like this comment is unrelated to this PR. This PR regards image conversion, not compression.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
rlm96: I think you're right, process a lot of files in the UI thread may be a hard job. Instead of that, I have applied another approach: after ```PickSingleFolderAsync()``` method as many tasks as images to be converted will be created, and then those tasks will be executed asynchronusly using ```Task.WhenAll``` [so we won´t be tie up a thread to just hanging](https://stackoverflow.com/questions/6123406/waitall-vs-whenall).

All tasks created [will run on a ThreadPool thread](https://docs.microsoft.com/en-us/dotnet/api/system.threading.threadpool?view=net-6.0#remarks) so we will use background threads for that.

It is worth highlighting that the user will see how ImageConversionWorkItems representation in the UI will be gradually phased out (due to ```work.DeleteCommand.Execute(null)```), so we don´t need to show a progress bar or something like that to show the user we're doing any kind of background job.

rlm96:
@btiteux I understand the recommendations and thoughts about the UI/UX you have provided. It is possible show to the user the extension related to the format choosen (for example, .jpg for JPEG format). My only concern is that if the user changes the format, all the extensions will change. Idk if this can be annoying from a user perspective, but if you think it won´t be I can do that change (@veler if you have some comments about this, please, I would love to read them).

In addition, I have implemented almost all changes, I just have to deal with the error handling from a UX perspective and see how we can show the error to the user in a friendly way 🤔.

veler:
> In addition, I have implemented almost all changes, I just have to deal with the error handling from a UX perspective and see how we can show the error to the user in a friendly way 🤔.

I can see 2 approaches:
1. If the conversion is done as soon as a file is selected (before clicking "Save All", like what PNG / JPEG Compressor does), we can show a button for each file item, like in PNG / JPEG Compressor. The button would pop up a content dialog with the error. Good thing about it is that it stays consistent with PNG Compressor in my opinion.
2. Otherwise, either we can popup a content dialog directly, or maybe show a banner like this below the Save All button:
image

What do you think?

rlm96:
> If the conversion is done as soon as a file is selected (before clicking "Save All", like what PNG / JPEG Compressor does), we can show a button for each file item, like in PNG / JPEG Compressor. The button would pop up a content dialog with the error. Good thing about it is that it stays consistent with PNG Compressor in my opinion.

@veler Conversion is done after clicking in **Save All** or **Save** buttons. So I would suggest use InfoBar.

veler:
> > If the conversion is done as soon as a file is selected (before clicking "Save All", like what PNG / JPEG Compressor does), we can show a button for each file item, like in PNG / JPEG Compressor. The button would pop up a content dialog with the error. Good thing about it is that it stays consistent with PNG Compressor in my opinion.
> 
> @veler Conversion is done after clicking in **Save All** or **Save** buttons. So I would suggest use InfoBar.

Let's do that then :)

rlm96:
@veler InfoBar approach applied 😄 I've fixed few things btw. All comments are (almost) completed 🥳 so I'm currently waiting for your final thoughts

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
```diff
@@ -0,0 +1,290 @@
...
+
+                    foreach (ImageConversionWorkItem work in works)
+                    {
+                        conversionTasks.Add(SaveConversionWorkItem(selectedFolder, work));
+                    }
+
+                    await Task.WhenAll(conversionTasks);
+                }
+            }
+            catch (Exception) { IsInfoBarOpen = true; }
```
veler, line 241: Let's log the exception, so users can share the issue when a problem occurs :).
https://github.com/veler/DevToys/blob/b8f9e4bb12eccf71da6dc2bc8479f0dd09d09138/src/dev/impl/DevToys/Core/Logger.cs#L60

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
```diff
@@ -0,0 +1,133 @@
...
+
+                DeleteCommand.Execute(this);
+            }
+        }
+
+        #endregion
+
+        private static void HandleError(Exception e)
+        {
+            throw e;
```
veler, line 109: I'm not following, where will this be catch?
Perhaps we should consider using `Logger.LogFaultAsync` instead.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
```diff
@@ -0,0 +1,290 @@
...
+        private void ExecuteCloseInfoBarButtonCommand()
+        {
+            IsInfoBarOpen = false;
+        }
+
+        #endregion
+
+        private async Task SaveConversionWorkItem(StorageFolder selectedFolder, ImageConversionWorkItem work)
+        {
+            StorageFile newFile = await selectedFolder.CreateFileAsync(work.FileName, CreationCollisionOption.ReplaceExisting);
```
veler, line 259: I tried to bug bash a little bit the feature and I think I found an issue here :)

Scenario :
1. Add a file `C:\folder\foo.jpeg` to convert to PNG.
2. Click `Save All` and save in `C:\folder`.
3. Line 259 crash because the file `foo.jpeg` is already opened.

I looked a little bit at where is the file opened, and it appears that it's due to the `ImageConversionWorkItem.Bitmap` property. That property is non-null at this moment and it seems like it keeps the file opened even is you correctly dispose the `IRandomAccessStream` in `DecodeImageAsync`.

In addition to this problem, I realized that the SoftwareBitmap is never disposed, meaning that we keep the bitmap indefinitely in memory (and therefore, memory leak). I just gave it a try and confirmed my thought:
1. Select 100 images in the tool.
2. Observe in Task Manager that the RAM increased a lot.
3. Click `Delete All` without doing the conversion.
4. All items disappear from the UI, but the RAM is still high in Task Manager.

Therefore, here is what I suggest.
1. Between the moment the user selects a file and the moment he clicks `Save`, minutes or even hours can elapse. Meanwhile, it seems like we don't need that SoftwareBitmap until the user Save the file. Therefore, as a first step, I'd suggest to refactor the code so DecodeImageAsync is only called in the Save/Conversion process.
2. As part of this refactoring,  DecodeImageAsync could return the SoftwareBitmap instead of using the Bitmap property. That would allow to have a `using (var bitmap = await DecodeImageAsync)`, so the resource can get disposed ASAP, and we can remove the `Bitmap` property.
3. Lines 261 to 268 are pretty much the same in ImageConversionWorkItem. Let's move this code in a new async method (off UI thread) in ImageConversionWorkItem so we avoid duplicating important piece of code.

Finally, I'm suspecting another issue with lines 259 and 260. Let's take the same scenario as above:
1. Add a file `C:\folder\foo.jpeg` to convert to PNG.
2. Click `Save All` and save in `C:\folder`.
Wouldn't line 259 **replace** `foo.jpeg` even if we don't write in the file? That doesn't sound right.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        private async Task DecodeImageAsync(StorageFile file)
+        {
+            await TaskScheduler.Default;
+
+            using (IRandomAccessStream stream = await file.OpenAsync(FileAccessMode.Read))
+            {
+                BitmapDecoder decoder = await BitmapDecoder.CreateAsync(stream);
+                var bitmap = await decoder.GetSoftwareBitmapAsync();
+                await ThreadHelper.RunOnUIThreadAsync(() => Bitmap = bitmap);
```
veler, line 129: As mentioned in my other comment, we never dispose `bitmap`, which leads to a memory leak.

veler (changes requested):
Thank you for this new update @rlm96 . I gave a try to the PR and overall it works great. However I found a memory leak and a bug that I would love to have it fixed before merging the PR. 

I think we're almost there :)

Please let me know if you need any help with the fix I proposed. :-)

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
rlm96: The purpose here is to raise an exception that wouldn´t be raised because it is being executed in a background (worker) thread, so it is possible to catch it in the ViewModel (ImageConversionWorkItem constructor is under a try). If you don´t think this approach will work, should we raise the exception in HandleError method using the UI thread? 

```
private async static void HandleError(Exception e)
{
      await ThreadHelper.RunOnUIThreadAsync(() => { throw e; });
}
```
or should we just log here the error?

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
rlm96: > Finally, I'm suspecting another issue with lines 259 and 260. Let's take the same scenario as above:
> 1. Add a file C:\folder\foo.jpeg to convert to PNG.
> 2. Click Save All and save in C:\folder.
> Wouldn't line 259 replace foo.jpeg even if we don't write in the file? That doesn't sound right.

About this, I've used the same approach that we're currently using in PngJpgCompressor tool. Anyway, answering your  comment I think it won´t be replaced because file extensions are different.

rlm96: On the other hand, about first issue:

> Between the moment the user selects a file and the moment he clicks Save, minutes or even hours can elapse. Meanwhile, it seems like we don't need that SoftwareBitmap until the user Save the file. Therefore, as a first step, I'd suggest to refactor the code so DecodeImageAsync is only called in the Save/Conversion process.

I think you're right, we're not diposing the Bitmap property but with the approach you've proposed we will need a StorageFile type property to decode the image when needed so will occur the same. My suggestion to fix point 2 and 3 is implement IDisposable in ```ImageConversionWorkItem``` class and dispose the Bitmap property if isn´t null. I've commited and pushed those changes so I would like to know your opinion 😄

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
veler: > The purpose here is to raise an exception that wouldn´t be raised because it is being executed in a background (worker) thread

It will still be raised, just not caught by anything. I'd suggest to just log the error in the Logger. :)

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
rlm96: > Scenario :
> 
> 1. Add a file C:\folder\foo.jpeg to convert to PNG.
> 2. Click Save All and save in C:\folder.
> 3. Line 259 crash because the file foo.jpeg is already opened.

Good catch! I have fixed this in [this commit](https://github.com/veler/DevToys/pull/139/commits/463854f3892b33d29a8391d5bf967172ea09f82a). This approach was also applied in JpgPngCompressor tool so it might be a good idea to review it, just in case.

veler: > > Finally, I'm suspecting another issue with lines 259 and 260. Let's take the same scenario as above:
> > 
> > 1. Add a file C:\folder\foo.jpeg to convert to PNG.
> > 2. Click Save All and save in C:\folder.
> >    Wouldn't line 259 replace foo.jpeg even if we don't write in the file? That doesn't sound right.
> 
> About this, I've used the same approach that we're currently using in PngJpgCompressor tool. Anyway, answering your comment I think it won´t be replaced because file extensions are different.

Can you please verify that it is the case? Because on line 259, we first create the file with the old name (foo.jpeg) at the location where the file exists already and we tell to replace the existing one. And only on the line after we rename it with a different extension. That is quite confusing to me.

> > Between the moment the user selects a file and the moment he clicks Save, minutes or even hours can elapse. Meanwhile, it seems like we don't need that SoftwareBitmap until the user Save the file. Therefore, as a first step, I'd suggest to refactor the code so DecodeImageAsync is only called in the Save/Conversion process.
> 
> I think you're right, we're not diposing the Bitmap property but with the approach you've proposed we will need a StorageFile type property to decode the image when needed so will occur the same. My suggestion to fix point 2 and 3 is implement IDisposable in `ImageConversionWorkItem` class and dispose the Bitmap property if isn´t null (before the garbage collection does that). I've commited and pushed those changes so I would like to know your opinion 😄

Yes, I've been thinking about using IDisposable, but there's still the fact that the bitmap will be loaded in memory for a while. I think it's best to load it the latest possible in this scenario, which is why I suggested my approach. :)

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
rlm96: Being disposed now so we don´t have to wait the garbage collector

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
rlm96: > Can you please verify that it is the case? Because on line 259, we first create the file with the old name (foo.jpeg) at the location where the file exists already and we tell to replace the existing one. And only on the line after we rename it with a different extension. That is quite confusing to me.

Just changed that hahaha

rlm96: > Yes, I've been thinking about using IDisposable, but there's still the fact that the bitmap will be loaded in memory for a while. I think it's best to load it the latest possible in this scenario, which is why I suggested my approach. :)

Both Bitmap or StorageFile are needed to be stored in the WorkItem as a property, so if we're not using the Bitmap property approach we will need to store the StorageFile as a property to call the decode method when saving.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
rlm96: > It will still be raised

Cool, so anyway error will be logged then in ViewModel

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
veler: > Both Bitmap or StorageFile are needed to be stored in the WorkItem as a property, so if we're not using the Bitmap property approach we will need to store the StorageFile as a property to call the decode method when saving 🤔.

That is fine. StorageFile is just a pointer to that file, it's not the actual file being loaded in memory. However. by keeping the StorageFile in the work item, we will need to check whether the file still exists on the hard drive when doing the conversion (after all, users could select a file in DevToys, delete the file from the hard drive, then click Save). :)

Feel free to jump on Discord for a quick chat if you want.

xeonmc:
I know it's a bit late in the PR but I think it is a nobrainer to add `.webp` support since it's supposed to be an all-encompassing format held back by the lack of "quick and dirty" tools to convert supported formats into them.

btiteux:
> I know it's a bit late in the PR but I think it is a nobrainer to add `.webp` support since it's supposed to be an all-encompassing format held back by the lack of "quick and dirty" tools to convert supported formats into them.

It's a good idea to add this type of image. 
@xeonmc can you create an issue to add this type? (with AVIF) 
This way we can finish this pull request.

Thank you

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
```diff
@@ -0,0 +1,118 @@
...
+
+            StorageFile? newFile = await savePicker.PickSaveFileAsync();
+
+            if (newFile is not null)
+            {
+                using (IRandomAccessStream outputStream = await newFile.OpenAsync(FileAccessMode.ReadWrite))
+                {
+                    BitmapEncoder? encoder = await ImageHelper.GetEncoderAsync(viewModel.ConvertedFormat, outputStream);
+
+                    var bitmap = await DecodeImageAsync();
```
veler, line 89: Don't forget to dispose the bitmap when you don't need it anymore 😉

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
```diff
@@ -0,0 +1,303 @@
...
+
+        #endregion
+
+        private async Task SaveConversionWorkItem(StorageFolder selectedFolder, ImageConversionWorkItem work)
+        {
+            StorageFile newFile = await selectedFolder.CreateFileAsync(string.Concat(work.File.DisplayName, ImageHelper.GetExtension(ConvertedFormat)), CreationCollisionOption.ReplaceExisting);
+            using (IRandomAccessStream outputStream = await newFile.OpenAsync(FileAccessMode.ReadWrite))
+            {
+                BitmapEncoder? encoder = await ImageHelper.GetEncoderAsync(ConvertedFormat, outputStream);
+                var bitmap = await work.DecodeImageAsync();
```
veler, line 276: Here too, please don't forget to dispose this resource after `FlushAsync`.

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
rlm96: Cool, so I have implemented that approach 😄

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
rlm96: I was about to push it 😝

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
rlm96: Same here, I was about to push it 😝

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConverterToolViewModel.cs
veler: Hahaha, sorry, I've been too fast then :-)

rlm96:
@veler applied all new commented changes 🥳

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ImageConverter/ImageConversionWorkItem.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            get => _fileSize;
+            private set
+            {
+                ThreadHelper.ThrowIfNotOnUIThread();
+                SetProperty(ref _fileSize, value);
+            }
+        }
+
+        internal string? ErrorMessage { get; private set; }
```
veler, line 38: This property doesn't seem to be used anymore. Can we remove it?

rlm96: Indeed, removed

veler (approved):
Overall, it looks good to me 🚀🎉
Thank you for your patience and thank you for your contribution! :D

@btiteux, can you please take a last look and merge if it looks good to you too?

Issue #168, opened 2022-01-20
Rounding error in JSON to YML conversion


beatbrot:
**Describe the bug**
In the JSON to YML converter, numbers are sometimes rounded wrongly.

**To Reproduce**

Convert the following JSON to YML:

```json
{
    "foo": 1.4
}
```

And you get:

```yml
foo: 1.3999999999999999
```

**Expected behavior**

```yml
foo: 1.4
```

**Configuration (please complete the following information):**
 - OS & Version: Windows 11 21H2
 - Theme: Dark
 - DevToys app version: 1.0.0.0 | X64 | RELEASE | ef783c8 | ef783c8

veler:
Hi,
Thanks for reporting this. We will try to fix it when we get some time. This repository notoriety blew up yesterday and we're taking sometime looking at all the feedback and triaging them :)

akarboush:
Is there really a need for a custom converter?
Why just not specify that floating point numbers should be parsed as `decimal` in the [JsonConvert.DeserializeObject](https://github.com/veler/DevToys/blob/3b612270587f89479e191fdddcab18ca4d08d501/src/dev/impl/DevToys/ViewModels/Tools/Converters/JsonYaml/JsonYamlToolViewModel.cs#L233)
```csharp
private readonly Newtonsoft.Json.JsonSerializerSettings _defaultJsonSerializerSettings = new () 
{
   FloatParseHandling = FloatParseHandling.Decimal 
};
..........
dynamic? jsonObject = JsonConvert.DeserializeObject<ExpandoObject>(input, _defaultJsonSerializerSettings );

```
Decimal would also handle a larger number than double

PR #201, opened 2022-01-22
Fixed rounding errors for JSON>YAML converter

sapvissers:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #168 

## What is the new behavior?

- Added a custom `DoubleConverter` to handle the correct double conversion from JSON to YAML, which fixed all rounding errors.

## Other information

I didn't implement the `ReadYaml()` method in the `DoubleConverter` because there is no use for it.

## Quality check


src/dev/impl/DevToys/ViewModels/Tools/Converters/JsonYaml/DoubleConverter.cs
```diff
@@ -0,0 +1,28 @@
...
+    internal class DoubleConverter : IYamlTypeConverter
+    {
+        public bool Accepts(Type type)
+        {
+            return type == typeof(double);
+        }
+
+        public object ReadYaml(IParser parser, Type type)
+        {
+            throw new NotImplementedException();
```
veler, line 19: What happens if the user converts from YAML to JSON? Would this method be called?

src/dev/DevToys.Startup/Package.appxmanifest
```diff
@@ -15,7 +15,7 @@
     <Identity
         Name="DevToysPreview"
         Publisher="CN=etiennebaudoux"
-        Version="0.0.0.0" />
+        Version="1.0.1.0" />
```
veler, line 18: Please revert this change.

src/dev/shared/SharedAssemblyInfo.cs
```diff
@@ -37,7 +37,7 @@
 
 // Please DO NOT commit changes on the 2 lines below unless
 // you're about to release a new version of the app on the Store.
-[assembly: AssemblyVersion("0.0.0.0")]
-[assembly: AssemblyFileVersion("0.0.0.0")]
+[assembly: AssemblyVersion("1.0.1.0")]
```
veler, line 40: Please revert this change too.

veler (commented):
Hello,

Thank you for making this fix. A couple of comments, but overall it looks good. :)

veler:
Additionally, I feel like this PR could fix #202, isn't it? Mind giving it a quick try?

akarboush:
Is there really a need for a custom converter?
Why just not specify that floating point numbers should be parsed as `decimal` in the [JsonConvert.DeserializeObject](https://github.com/veler/DevToys/blob/3b612270587f89479e191fdddcab18ca4d08d501/src/dev/impl/DevToys/ViewModels/Tools/Converters/JsonYaml/JsonYamlToolViewModel.cs#L233)
```csharp
private readonly Newtonsoft.Json.JsonSerializerSettings _defaultJsonSerializerSettings = new () 
{
   FloatParseHandling = FloatParseHandling.Decimal 
};
..........
dynamic? jsonObject = JsonConvert.DeserializeObject<ExpandoObject>(input, _defaultJsonSerializerSettings );

```
Decimal would also handle a larger number than double

sapvissers:
> Is there really a need for a custom converter? Why just not specify that floating point numbers should be parsed as `decimal` in the [JsonConvert.DeserializeObject](https://github.com/veler/DevToys/blob/3b612270587f89479e191fdddcab18ca4d08d501/src/dev/impl/DevToys/ViewModels/Tools/Converters/JsonYaml/JsonYamlToolViewModel.cs#L233)
> 
> ```cs
> private readonly Newtonsoft.Json.JsonSerializerSettings _defaultJsonSerializerSettings = new () 
> {
>    FloatParseHandling = FloatParseHandling.Decimal 
> };
> ..........
> dynamic? jsonObject = JsonConvert.DeserializeObject<ExpandoObject>(input, _defaultJsonSerializerSettings );
> ```
> 
> Decimal would also handle a larger number than double

Thank you for this feedback, didn't know this was possible!
I've refactored the code as you sugested and it works great.

sapvissers:
> Additionally, I feel like this PR could fix #202, isn't it? Mind giving it a quick try?

This PR doesn't fix #202, I did look at it briefly, but couldn't find a quick solution so I'll look into that issue when I have the time.

veler:
That looks great! Thank you very much for fixing this 🚀😁

Issue #141, opened 2022-01-19
Checksum generator


veler:
**Is your feature request related to a problem? Please describe.**
As a user, I'd like to be able to calculate the checksum of a file.

**Describe the solution you'd like**
Let's add a tool for generating Checksum.

Joaolfelicio:
Hey @veler , this seems interesting and maybe I could help.
Can you give more information about this feature?

veler:
Hello @Joaolfelicio , sure thing!
The workflow behind this tool would be:
1. User selects 1 file.
2. User choose whether an MD5, SHA-1, SHA-256...etc hash should be generated.
3. When computing the hash, ideally a progress bar would be displayed (or at minimum a UI indicating that a background process is in progress).
4. The resulted hash is displayed on screen.
5. Finally, perhaps an extra field would allow the user to compare the generated hash with a hash of his choice.

We already have a Hash Generator for text. What I'm not very sure is whether it would make sense to support both (text and file hash generation) in 1 tool, or if we should split it in 2 tools.

Regardless, you can find some guidance on the creation of a new tool here: https://github.com/veler/DevToys/blob/main/CONTRIBUTING.md

You can also take some existing tool as a template. I'd recommend to look at the implementation of the existing Hash Generator tool and Color Blindness Simulator (freshly added to the app but not published yet: https://github.com/veler/DevToys/pull/138)

Let me know if this help or if you want to brainstorm it :)

Joaolfelicio:
Sounds good to me!
If you want, you can assign this to me.

I think if we separate the tools, it would be a bit more clear of its purpose, even if the insides are pretty similar. `Checksum File Tool` would be pretty verbose about its intent.
Let me know what you think!

veler:
Agreed! A separated tool is probably best. Go for it! 😁 😁 😁

jwfxpr:
Bug report at #186 is relevant to this discussion. 😊

PR #234, opened 2022-01-24
Feature/check sum generator

Joaolfelicio:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

As a user, I'd like to be able to calculate the checksum of a file and compare it with a hash value so I'm able to know if the file was modified.

Issue Number: #141

## What is the new behavior?

As a user, I'd like to be able to calculate the checksum of a file using multiple hashing algorithms like MD5, SHA1, SHA256, SHA384 or SHA512.

It will display a loading spinner while a file is being hashed unless the file is small, so we don't see the load flicking.

## Other information
main-page

check-sum

check-sum-hashing-algos

check-sum-loading

check-sum-output

## Quality check

## What's missing

1. Not sure which font icon/svg to use, open to suggestions!
2. For the moment, the control output comparer is not doing anything, not sure if we want to keep it or what would be the best UX for this control.
3. I did realize that the compact overlay button on the top sometimes throws, could be local as no change related to it was performed.

I will need some help with the above points, thanks.


Joaolfelicio:
Regarding point number 3, seems like it's only in debug mode!

src/dev/impl/DevToys/ViewModels/Tools/Generators/CheckSumGenerator/CheckSumGeneratorToolViewModel.cs
```diff
@@ -0,0 +1,429 @@
...
+                using HashAlgorithm? hashAlgo = CreateHashAlgorithm(inputHashingAlgorithm);
+                Stream? fileStream = await inputFile.OpenStreamForReadAsync();
+
+                ShouldDisplayProgress = HashingHelper.ComputeHashIterations(fileStream) > _displayAboveIterations;
+
+                byte[]? fileHash = await HashingHelper.ComputeHashAsync(
+                        hashAlgo,
+                        fileStream,
+                        new Progress<HashingProgress>(UpdateProgress),
+                        _cancellationTokenSource.Token);
```
btiteux, line 373: I see that you use a Stream but I don't see the dispose of this stream?

src/dev/impl/DevToys/Models/HashingAlgorithmDisplayPair.cs
```diff
@@ -0,0 +1,34 @@
...
+        private HashingAlgorithmDisplayPair(string displayName, HashingAlgorithm value)
+        {
+            DisplayName = displayName;
+            Value = value;
+        }
+
+        public bool Equals(HashingAlgorithmDisplayPair other) => other.Value == Value;
+    }
+
+    public enum HashingAlgorithm
```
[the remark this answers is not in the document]
Joaolfelicio: Done

src/dev/impl/DevToys/ViewModels/Tools/Generators/CheckSumGenerator/CheckSumGeneratorToolViewModel.cs
Joaolfelicio: Well spotted, missed that one 😁

src/dev/impl/DevToys/Helpers/HashingHelper.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Threading.Tasks;
+using DevToys.Models;
+using DevToys.Shared.Core;
+
+namespace DevToys.Helpers
+{
+    internal static class HashingHelper
+    {
+        internal static async Task<byte[]> ComputeHashAsync(
+        HashAlgorithm hashAlgorithm,
```
veler, line 14: Please indent the parameters by adding 4 spaces.

src/dev/impl/DevToys/Helpers/HashingHelper.cs
```diff
@@ -0,0 +1,70 @@
...
+        Stream stream,
+        IProgress<HashingProgress> progress,
+        CancellationToken cancellationToken = default,
+        int bufferSize = 1024 * 1024)
+        {
+            Assumes.NotNull(stream, nameof(stream));
```
veler, line 20: Can you please use `Arguments.NotNull` instead? :)

src/dev/impl/DevToys/Helpers/HashingHelper.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace DevToys.Helpers
+{
+    internal static class HashingHelper
+    {
+        internal static async Task<byte[]> ComputeHashAsync(
+        HashAlgorithm hashAlgorithm,
+        Stream stream,
+        IProgress<HashingProgress> progress,
+        CancellationToken cancellationToken = default,
```
veler, line 17: Ideally, a `CancellationToken` should never be an optional argument. Making it optional encourage future maintainer at bad practice by omitting it.

src/dev/impl/DevToys/Helpers/HashingHelper.cs
```diff
@@ -0,0 +1,70 @@
...
+
+                progress.Report(new HashingProgress(stream.Length, totalBytesRead));
+                cancellationToken.ThrowIfCancellationRequested();
+            }
+            return hashAlgorithm.Hash ?? Array.Empty<byte>();
+        }
+
+        internal static int ComputeHashIterations(Stream stream, int bufferSize = 1024 * 1024)
+        {
+            Assumes.NotNull(stream, nameof(stream));
```
veler, line 55: Sam here

src/dev/impl/DevToys/Helpers/HashingHelper.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
```
veler, line 1: Please add `#nullable enable` at the top of the file.

src/dev/impl/DevToys/ViewModels/Tools/Generators/CheckSumGenerator/CheckSumGeneratorToolViewModel.cs
```diff
@@ -0,0 +1,429 @@
...
+            {
+                Output = Output!.ToLowerInvariant();
+            }
+        }
+
+        private void CheckSum()
+        {
+            lock (_lockObject)
+            {
+                ResetProgress();
```
veler, line 325: I played a little bit with the tool and found an issue in the following scenario:
1. Select a large file.
2. The hash calculation is in progress, the progress ring is increasing. While this time, select another file.

When doing this, the previous calculation process isn't canceled. You might be able to fix it by calling invoking the `CancelCommand` instread of `ResetProgress`.

src/dev/impl/DevToys/ViewModels/Tools/Generators/CheckSumGenerator/CheckSumGeneratorToolViewModel.cs
```diff
@@ -0,0 +1,429 @@
...
+            if (_isCalculationInProgress || InputFile is null)
+            {
+                return;
+            }
+
+            _isCalculationInProgress = true;
+
+            string? hashOutput = await CalculateFileHash(InputHashingAlgorithm.Value, InputFile);
+
+            ThreadHelper.RunOnUIThreadAsync(() =>
```
veler, line 341: Await this method.

src/dev/impl/DevToys/ViewModels/Tools/Generators/CheckSumGenerator/CheckSumGeneratorToolViewModel.cs
```diff
@@ -0,0 +1,429 @@
...
+
+        private async Task CheckSumAsync()
+        {
+            if (_isCalculationInProgress || InputFile is null)
+            {
+                return;
+            }
+
+            _isCalculationInProgress = true;
+
```
veler, line 338: Please call `await TaskScheduler.Default;` here. This allows to force the method to go out of the UI thread, which guarantees us to not hanging the UI until we call RunOnUIThreadAsync

src/dev/impl/DevToys/ViewModels/Tools/Graphic/ColorBlindnessSimulator/ColorBlindnessSimulatorToolViewModel.cs
```diff
@@ -52,10 +52,10 @@ public sealed class ColorBlindnessSimulatorToolViewModel : ObservableRecipient,
 
         internal ColorBlindnessSimulatorStrings Strings => LanguageManager.Instance.ColorBlindnessSimulator;
 
-        internal bool IsSelectFilesAreaHighlithed
+        internal bool IsSelectFilesAreaHighlighted
```
veler, line 55: Good catch! Thank you!

src/dev/impl/DevToys/Views/Tools/Generators/CheckSumGenerator/CheckSumGeneratorToolPage.xaml
```diff
@@ -0,0 +1,172 @@
...
+            IsReadOnly="True"
+            Text="{x:Bind ViewModel.Output, Mode=OneWay}"/>
+
+        <controls:CustomTextBox
+            Header="{x:Bind ViewModel.Strings.OutputComparer}"
+            Height="50"
+            AcceptsReturn="True"
+            Text="{x:Bind ViewModel.OutputComparer, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}" />
+
+        <controlsToolKit:Loading x:Name="LoadingControl" IsLoading="{x:Bind ViewModel.ShouldDisplayProgress}"  >
```
veler, line 162: it seems like this isn't used. Should we remove it?

src/dev/shared/DevToys.Shared/Core/Assumes.cs
```diff
@@ -35,5 +35,13 @@ public static void IsTrue(bool value, string argumentName)
                 throw new Exception($"'{argumentName}' is expected to be True");
             }
         }
+
+        public static void NotZeroOrBelow(long value, string argumentName)
```
veler, line 39: If I'm not wrong, you wanted to use this method to check an argument. If so, can you please move this method to `Arguments` class instead?

veler (changes requested):
Hello,
Thank you very much for this contribution! Overall, it looks very good. Just a few details here and there. :-)

src/dev/impl/DevToys/Helpers/HashingHelper.cs
Joaolfelicio: Done

src/dev/impl/DevToys/Helpers/HashingHelper.cs
Joaolfelicio: Good point, removed the default value.

src/dev/impl/DevToys/Helpers/HashingHelper.cs
Joaolfelicio: Done

src/dev/impl/DevToys/Helpers/HashingHelper.cs
Joaolfelicio: Missed this one!

src/dev/impl/DevToys/ViewModels/Tools/Generators/CheckSumGenerator/CheckSumGeneratorToolViewModel.cs
Joaolfelicio: Good catch. Initially was invoking the `CancelCommand` on each calculation, but it was messing up the logic to revert back the dropdown to the previous hashing algorithm if you cancel the calculation.
I'm now passing an argument that specifies if we should revert to the previous hash algorithm or not.
It should be fixed now.

src/dev/impl/DevToys/ViewModels/Tools/Generators/CheckSumGenerator/CheckSumGeneratorToolViewModel.cs
Joaolfelicio: Done.

src/dev/impl/DevToys/ViewModels/Tools/Generators/CheckSumGenerator/CheckSumGeneratorToolViewModel.cs
Joaolfelicio: Good point, added it!

src/dev/impl/DevToys/Views/Tools/Generators/CheckSumGenerator/CheckSumGeneratorToolPage.xaml
Joaolfelicio: Whoopsie! 😅
Removed the control and the import statement.

src/dev/shared/DevToys.Shared/Core/Assumes.cs
Joaolfelicio: Done.

veler (approved):
Awesome! Thank you for the update! It looks good to me. @btiteux , can you please take a look and merge if it looks good to you too?

Joaolfelicio:
Nice! Could you please just take a look at points number 1 and 2?

veler:
> * Not sure which font icon/svg to use, open to suggestions!
> * For the moment, the control output comparer is not doing anything, not sure if we want to keep it or what would be the best UX for this control.

Oh sorry I missed that.

> Not sure which font icon/svg to use, open to suggestions!

It looks good to me. Worst case we can change that later.

> For the moment, the control output comparer is not doing anything, not sure if we want to keep it or what would be the best UX for this control.

That's fine too :)

src/dev/impl/DevToys/ViewModels/Tools/Generators/CheckSumGenerator/CheckSumGeneratorToolViewModel.cs
```diff
@@ -0,0 +1,432 @@
...
+            await ThreadHelper.RunOnUIThreadAsync(() =>
+            {
+                if (hashOutput != null)
+                {
+                    Output = hashOutput;
+                    _outputHashingAlgorithm = InputHashingAlgorithm;
+                }
+
+                _isCalculationInProgress = false;
+                ShouldDisplayProgress = false;
```
veler, line 356: I'm facing a race condition on my machine and I think that this line along with 343 and maybe a few others are the cause.

The scenario I have is:
1. Select a file. Computation start. Progress bar is displayed.
2. Select another file while the computation of the 1st one isn't done yet.
3. In this scenario, the progress bar simply disappear.

I think the race here is that line 343 (with the new file) runs before 356 (with the old file) is executed.

A proper fix is probably to cache the task on line 326 and await it (when not null) right after line 339. Thoughts?

Issue #239, opened 2022-01-24
HashGenerator output on base64


andrescirulo:
In some cases it is useful that the output of a hash calculation is in base64 instead of hexadecimal. It would be good to have the option to select the type of output.

PR #240, opened 2022-01-24
HashGenerator output on base64 or hex

andrescirulo:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The output of a hash generation is always Hex. Sometimes is useful for the output to be in base64.

Issue Number: #239 

## What is the new behavior?

- An option to select the Output Type between Hex and Base64 has been added

### DIFF
--- a/src/dev/impl/DevToys/ViewModels/Tools/Generators/HashGenerator/HashGeneratorToolViewModel.cs
+++ b/src/dev/impl/DevToys/ViewModels/Tools/Generators/HashGenerator/HashGeneratorToolViewModel.cs
@@ -50,7 +50,9 @@ namespace DevToys.ViewModels.Tools.HashGenerator
         private string? _sha1;
         private string? _sha256;
         private string? _sha512;
-        private const string DefaultOutputType = "Hex";
+        private const string HexOutput = "Hex";
+        private const string Base64Output = "Base64";
+        private const string DefaultOutputType = HexOutput;
 
         public Type View { get; } = typeof(HashGeneratorToolPage);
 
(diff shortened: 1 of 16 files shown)


src/dev/impl/DevToys/ViewModels/Tools/Generators/HashGenerator/HashGeneratorToolViewModel.cs
```diff
@@ -161,7 +185,15 @@ private async Task<string> CalculateHashAsync(string alrogithmName, string text)
                 IBuffer buffer = CryptographicBuffer.ConvertStringToBinary(text, BinaryStringEncoding.Utf8);
                 buffer = algorithmProvider.HashData(buffer);
 
-                string? hash = CryptographicBuffer.EncodeToHexString(buffer);
```
btiteux, line 164: Hello, 
I see that you compare two different values, can you use a String.Equals to do that?
And also please avoid using string directly here and define an internal for example, you can see and example in the "JsonYamlToolViewModel"
Thank you

andrescirulo: I commited the changes, please check if now it's better

btiteux: It's seems good to me @veler do you want to check too?

veler (approved):
Looking good, thanks for this contribution :)

PR #260, opened 2022-01-26
Add a copy button to the MarkdownPreview tool to copy the HTML output

praneetloke:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

There is no way to copy the HTML output from the markdown preview.

Issue Number: N/A

## What is the new behavior?

A new copy button above the `Webview` control allows the user to copy the contents of the `markdown-body` element.

<details>
<summary>Screenshot</summary>

Screenshot 2022-01-25 181200

</details>

## Other information

I am a newbie at UWP app development and especially with the MVVM pattern.

## Quality check

  - I could not get this to work on my machine for some reason even though my changes should not affect the Release build config. I was getting an error about a `SetVersion` task being unable to run because `Microsoft.Build.Utilities` version `3.5.0.0` wasn't available despite having a successful Nuget restore.

### DIFF
--- a/src/dev/impl/DevToys/Views/Tools/Text/MarkdownPreview/MarkdownPreviewToolPage.xaml.cs
+++ b/src/dev/impl/DevToys/Views/Tools/Text/MarkdownPreview/MarkdownPreviewToolPage.xaml.cs
@@ -1,13 +1,17 @@
 ﻿#nullable enable
 
+using System;
 using DevToys.Api.Core.Navigation;
+using DevToys.Core;
 using DevToys.Messages;
 using DevToys.Shared.Core;
 using DevToys.ViewModels.Tools.MarkdownPreview;
 using Microsoft.Toolkit.Mvvm.Messaging;
+using Windows.ApplicationModel.DataTransfer;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 using Windows.UI.Xaml.Navigation;
+using Clipboard = Windows.ApplicationModel.DataTransfer.Clipboard;
 
 namespace DevToys.Views.Tools.MarkdownPreview
 {
@@ -64,6 +68,26 @@ namespace DevToys.Views.Tools.MarkdownPreview
         {
             Arguments.NotNull(message, nameof(message));
             PreviewWebView.NavigateToString(message.Html);
+        }
+
+        private async void CopyButton_Click(object sender, RoutedEventArgs e)
+        {
+            try
+            {
+                var data = new DataPackage
+                {
+                    RequestedOperation = DataPackageOperation.Copy
+                };
+                string markdownBodyEl = "document.getElementsByClassName('markdown-body')[0].innerHTML";
+                data.SetText(await PreviewWebView.InvokeScriptAsync("eval", new string[] { markdownBodyEl }) ?? string.Empty);
+
+                Clipboard.SetContentWithOptions(data, new ClipboardContentOptions() { IsAllowedInHistory = true, IsRoamable = true });
+                Clipboard.Flush();
+            }
+            catch (Exception ex)
+            {
+                Logger.LogFault("Failed to copy from webview", ex);
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/dev/impl/DevToys/Views/Tools/Text/MarkdownPreview/MarkdownPreviewToolPage.xaml.cs
```diff
@@ -65,5 +69,25 @@ public void Receive(NavigateToMarkdownPreviewHtmlMessage message)
             Arguments.NotNull(message, nameof(message));
             PreviewWebView.NavigateToString(message.Html);
         }
+
+        private async void CopyButton_Click(object sender, RoutedEventArgs e)
```
praneetloke, line 73: * I think by MVVM convention this is probably an anti-pattern? That is, to have the click handler in the view?
* Also it feels a little silly to have to extract the HTML body from the webview rather than having access to the converted HTML. An alternative I considered was to make that property accessible to the view instead of doing what I did here. However, I wasn't sure of the convention of accessing the ViewModel's property from the View?

src/dev/impl/DevToys/Views/Tools/Text/MarkdownPreview/MarkdownPreviewToolPage.xaml.cs
```diff
@@ -65,5 +69,25 @@ public void Receive(NavigateToMarkdownPreviewHtmlMessage message)
...
+        {
+            try
+            {
+                var data = new DataPackage
+                {
+                    RequestedOperation = DataPackageOperation.Copy
+                };
+                string markdownBodyEl = "document.getElementsByClassName('markdown-body')[0].innerHTML";
```
praneetloke, line 81: I have to admit that this whole method is mostly an exact copy of similar functionality I found in the `CodeEditor` control class except for the strategy of extracting the HTML body from the webview. Speaking of which, is there a better way to access it? (somewhat related to my second point in my [comment](https://github.com/veler/DevToys/pull/260/files#r792285172) above.)

src/dev/impl/DevToys/Views/Tools/Text/MarkdownPreview/MarkdownPreviewToolPage.xaml.cs
veler: Hello,

Thanks for your contribution. 
Yes, this is an anti-pattern. The approach with MVVM is to make a command. In short, the button would be directly linked to something in the view model.
As an example, you can look at UUID Generator's "GenerateCommand", which is called when clicking the "Generate" button. https://github.com/veler/DevToys/blob/e2c9b71afd91c3f8135ea4bb8480c924773a3782/src/dev/impl/DevToys/ViewModels/Tools/Generators/GuidGenerator/GuidGeneratorToolViewModel.cs#L152-L152

You can look in the corresponding XAML how it is used.

For retrieving the HTML to copy, my advice is to generate the HTML itself, without trying to access the WebView (best practice is that the ViewModel doesn't interact directly with the controls in the UI).
For this, you can create a private method with this code:
https://github.com/veler/DevToys/blob/e2c9b71afd91c3f8135ea4bb8480c924773a3782/src/dev/impl/DevToys/ViewModels/Tools/Text/MarkdownPreview/MarkdownPreviewToolViewModel.cs#L137-L138
And pass the property `InputValue` in argument. Once you get the result of this method, you can copy it to the clipboard.
It will give you the exact same HTML :)

src/dev/impl/DevToys/Views/Tools/Text/MarkdownPreview/MarkdownPreviewToolPage.xaml.cs
veler: See my previous comment :)

src/dev/impl/DevToys/Views/Tools/Text/MarkdownPreview/MarkdownPreviewToolPage.xaml.cs
praneetloke: @veler thanks for the info. Let me give that a shot. It'll be a good learning experience for me.

praneetloke:
@veler I've updated the PR. Can you please take another look?

veler:
Awesome! It looks all good :) Thank you for this!

Issue #136, opened 2022-01-05
SQL formatter


rlm96:
**Is your feature request related to a problem? Please describe.**
As a software engineer, I use SQL formatter to make my queries more understandable. This can be done manually or using online tools but sometimes due to data restriction policies this is not possible.

**Describe the solution you'd like**
I think it would be a good idea to implement a SQL formatter.

**Describe alternatives you've considered**
Online tools such us https://www.dpriver.com/pp/sqlformat.htm or manually.

**Additional context**
N/A

veler:
Hi,

I love this idea. :) Definitely something we can add! Thank you for suggesting this.
We're planning to release DevToys 1.0 in January. We frozen feature work to focus on translation and bug fixes at the moment. Once 1.0 is release, we can look at including the formatter.
Meanwhile, even if 1.0 isn't released yet, we can still commit to `main` branch, so feel free to contribute if you wish. Otherwise, we will look at implementing it in a few weeks probably.
Thank you again!

veler:
This JavaScript library under MIT license seems to be one of the best open source SQL formatter out there.
https://github.com/zeroturnaround/sql-formatter

It may be interesting to make a fork of it in C#.

vukovinski:
This one is build in WPF, and already used in files.community

https://github.com/QL-Win/QuickLook

SamNHS:
I've tried the zeroturnaround formatter and it doesn't work so well with MSSQL, especially when using SQLCMD or @@
```sql
SET ANSI_NULLS, ANSI_PADDING, ANSI_WARNINGS, ARITHABORT, CONCAT_NULL_YIELDS_NULL, QUOTED_IDENTIFIER ON;

SET NUMERIC_ROUNDABORT OFF;


GO
:setvar DatabaseName "dbname"

GO
:on error exit
GO
/*
Detect SQLCMD mode and disable script execution if SQLCMD mode is not supported.
To re-enable the script after enabling SQLCMD mode, execute the following:
SET NOEXEC OFF; 
*/
:setvar __IsSqlCmdEnabled "True"
GO
IF N'$(__IsSqlCmdEnabled)' NOT LIKE N'True'
           BEGIN
     PRINT N'SQLCMD mode must be enabled to successfully execute this script.';
   SET NOEXEC ON;
 END


GO
USE [$(DatabaseName)];

/* INSERT, ALTER transactions here */

IF @@ERROR <> 0
                  AND @@TRANCOUNT > 0
    BEGIN
[...]

PR #274, opened 2022-01-27
Added SQL Formatter tool

veler:
## Pull request type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

As a developer working with database, I'd like to be able to format some queries without having to paste them on a random website.

Issue Number: #136 

## What is the new behavior?

Added a SQL Formatter based on https://github.com/zeroturnaround/sql-formatter

It isn't perfect though. This is a literal translation of the JS library. Some optimizations can be made to reduce the allocation of strings.

## Other information

image

image

image

image

## Quality check


veler:
@akarboush, please feel free to take a look at it :)

akarboush:
@veler thanks for the mention.
I took a brief look at it, and It looks similar to mine since it's a direct translation. 
As you already mentioned, there's a lot of optimization to do (string allocation and some regex could be replaced with a plain C# methods too).

For the mean time, I'll try to optimize it on my branch and then propose the changes if there aren't any here yet

src/dev/impl/DevToys/Helpers/SqlFormatter/Core/Formatter.cs
```diff
@@ -0,0 +1,347 @@
...
+        {
+            string formattedQuery = string.Empty;
+
+            Assumes.NotNull(_tokens, nameof(_tokens));
+            for (int i = 0; i < _tokens!.Count; i++)
+            {
+                _index = i;
+
+                Token token = TokenOverride(_tokens[i]);
+                if (token.Type == TokenType.LineComment)
```
Joaolfelicio, line 91: We could clean a bit all the if else's by using pattern matching.

veler:
> @veler thanks for the mention. I took a brief look at it, and It looks similar to mine since it's a direct translation. As you already mentioned, there's a lot of optimization to do (string allocation and some regex could be replaced with a plain C# methods too).
> 
> For the mean time, I'll try to optimize it on my branch and then propose the changes if there aren't any here yet

Sounds good. I think I will do a little bit of clean up before merging, but likely no optimization yet. While all these regex and strings are suboptimal, it seems to work relatively well as long as the SQL query isn't super long.

I'd suggest to do the optimizations in a separated PR, what do you think @akarboush ?

PR #299, opened 2022-01-29
Number converter fields as inputs

### LINKED ISSUE #256: Allow editing numbers in the output fields of Number Base Converter
### What's the Problem?

the current input -> output UI layout is great for the use case of having an unknown string of numbers and wanting to find out in which number base does it have a special pattern, by going through each of the options in the dropdown

However, if I already know what base my number is in, it is much faster if I could just click on one of the output fields and enter it there and have the other fields update automatically, including the input field. 

This is also useful for if I want to alter the pattern of numbers after finding the right source base using the tool.

Example: 

1. I found a string "157255" somewhere representing a magic number. 
2. I put it into the converter and go through the options, and find that if "157255" is in octal, then it will say "DEAD" in hexadecimal. 
3. Now I'm wondering what "DEED" would look like in the original octal source, so naturally I click on the hexadecimal output field to change the "A" into an "E"
4. Oops, I can't edit the output field as the app is currently. So instead I have to do the following steps:
   1. note and remember that the magic number base is hexadecimal
   2. copy the string "DEAD" from the hex output field (the string could've been long and thus infeasible to re-type by hand
   3. recall from experience that I need to change the dropdown first before pasting, otherwise it gets messed up (this auto-converting behaviour is good, it only becomes awkward in an in-place reconversion workaround like this use case)
   4. change the dropdown to hex
   5. paste the copied string
   6. change the letter "A" into "E"
  
Contrast that with doing an in-place edit of highlighting "A" and then pressing E, this is quite cumbersome and the very user-friendly design of the dropdown-autoconvert behaviour ended up getting in the way.

### Solution/Idea

allow in-place edits in the output fields

youkai95:
Closes #256 but with new issue lol
At the other hand, in #256 there are some features mentioned that could conflict with these new changes. The Input field now is removed, so if you want any custom number base input, then we will need to change the interface and how it works.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

You could change the numbers using any output field. Now there are some changes in the UI, as some options (number format) and Input field now become deprecated.

Issue Number: #256 

## What is the new behavior?

- You can use any output field as an input.

## Other information

~~Now there is a new issue... I am not sure why is happening that when the input is reformatted the text cursor returns to the left of the input field. I could keep an eye on it and try to get it solved.~~
Ok, this issue was solved with a workaround. Now the field where you decide to write the input doesn't receive any format (which could be considered as another issue)... TextBox will always reset the cursor position when updating the content, so it could be very cumbersome to design a reliable solution where cursor does not lose the position when the input should be formatted.

## Quality check

### DIFF
--- a/src/dev/impl/DevToys/ViewModels/Tools/Converters/NumberBaseConverter/NumberBaseFormatter.cs
+++ b/src/dev/impl/DevToys/ViewModels/Tools/Converters/NumberBaseConverter/NumberBaseFormatter.cs
@@ -141,6 +141,10 @@ namespace DevToys.ViewModels.Tools.Converters.NumberBaseConverter
         private static string FormatNumber(ReadOnlySpan<char> buffer, NumberBaseFormat baseNumber, bool isFormatted, int index)
         {
             var builder = new StringBuilder();
+            if (buffer[index - 1] == '-')
+            {
+                builder.Append(buffer[--index]);
+            }
 
             for (int builderIndex = --index; builderIndex >= 0; builderIndex--)
             {
--- a/src/dev/impl/DevToys/ViewModels/Tools/Converters/NumberBaseConverter/NumberBaseFormatter.cs
+++ b/src/dev/impl/DevToys/ViewModels/Tools/Converters/NumberBaseConverter/NumberBaseFormatter.cs
@@ -122,6 +122,24 @@ namespace DevToys.ViewModels.Tools.Converters.NumberBaseConverter
                 }
             }
 
+            return FormatNumber(buffer, baseNumber, isFormatted, index);
+        }
+
+        /// <summary>
+        /// Format <paramref name="number"/> based on <paramref name="baseNumber"/> format definition 
+        /// </summary>
+        /// <param name="number">String representation of the number</param>
+        /// <param name="baseNumber">Current base number <see cref="NumberBaseFormat"/></param>
+        /// <returns>Formatted number based on <paramref name="baseNumber"/> format definition</returns>
+        public static string FormatNumber(string number, NumberBaseFormat baseNumber)
+        {
+            char[] charArray = RemoveFormatting(number).ToCharArray();
+            Array.Reverse(charArray);
+            return FormatNumber(charArray, baseNumber, true, charArray.Length);
+        }
+
+        private static string FormatNumber(ReadOnlySpan<char> buffer, NumberBaseFormat baseNumber, bool isFormatted, int index)
+        {
             var builder = new StringBuilder();
 
             for (int builderIndex = --index; builderIndex >= 0; builderIndex--)
--- a/src/dev/impl/DevToys/Helpers/NumberBaseHelper.cs
+++ b/src/dev/impl/DevToys/Helpers/NumberBaseHelper.cs
@@ -16,13 +16,7 @@ namespace DevToys.Helpers
                 return false;
             }
 
-            long.TryParse(input, NumberStyles.HexNumber, null, out long output);
-            if (output is not 0)
-            {
-                return true;
-            }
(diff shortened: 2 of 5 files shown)


btiteux:
Hello,
Thank for the contribution.

I see one problem (that I didn't handle in the original version) we don't have the formatting in the "input" field.

For example, if I use the Hexadecimal field and have the Format Number switch to on. 

I should have:
`1B5 BDC3 B7D6 97B3`
instead of:
`1B5BDC3B7D697B3`

Same for other fields.

youkai95:
Yes, I know that is happening and I stated it in the "Other Information" section of the PR.
> Now there is a new issue... I am not sure why is happening that when the input is reformatted the text cursor returns to the left of the input field. I could keep an eye on it and try to get it solved.
 Ok, this issue was solved with a workaround. Now the field where you decide to write the input doesn't receive any format (which could be considered as another issue)... TextBox will always reset the cursor position when updating the content, so it could be very cumbersome to design a reliable solution where cursor does not lose the position when the input should be formatted.

I have an idea to allow formatting input but it would be a pickaxe solution xD

xeonmc:
> TextBox will always reset the cursor position when updating the content, so it could be very cumbersome to design a reliable solution where cursor does not lose the position when the input should be formatted.

why not just defer formatting update until cursor focus exits the input box?

youkai95:
> why not just defer formatting update until cursor focus exits the input box?

Thats a nice solution. I was thinking on hot formatting but that could be confusing to the user, even with the guarantee that the cursor will never lose the correct position.

I will push soon that fix, just give me a while as today is my sister birthday 😆

src/dev/impl/DevToys/ViewModels/Tools/Converters/NumberBaseConverter/NumberBaseFormatter.cs
```diff
@@ -122,7 +122,29 @@ public static string LongToBase(long number, NumberBaseFormat baseNumber, bool i
...
+            Array.Reverse(charArray);
+            return FormatNumber(charArray, baseNumber, true, charArray.Length);
+        }
+
+        private static string FormatNumber(ReadOnlySpan<char> buffer, NumberBaseFormat baseNumber, bool isFormatted, int index)
+        {
             var builder = new StringBuilder();
+            if (buffer[index - 1] == '-')
+            {
+                builder.Append(buffer[--index]);
```
youkai95, line 146: I included this fix here because otherwise will be merge conflicts if I push it in other PR.

veler: Oh that's great! Thanks for fixing it!
Actually, would you be able to add a unit test for it, please? We already have some in `BaseNumberFormatterTests` class.

youkai95:
About this fix for formatting I must admit that I am not happy with the result... It works and it is open to further features, however I feel there are many unnecessary things happening behind and some extra code is needed to extend this tool.

xeonmc:
Does this PR defer formatting of just the focused field or does it defer formatting of all fields in the document?

youkai95:
> Does this PR defer formatting of just the currently focused field, or does it defer formatting of all fields on the page until a focus change event is triggered? The former behaviour was what was meant in my prior suggestion, though upon rereading I realized that it might not have been expressed clearly enough.

It just defer formatting of the field that lose focus but I used the same event for every field. Fields without focus keeps formatting as the input changes.

veler:
Hello, sorry for the very late review. It looks all good now :) I will merge it. Thank you very much for this improvement!!

veler (approved):

PR #320, opened 2022-02-01
SOLID design for format converter tool

### LINKED ISSUE #247: SOLID suggerence for tools
## General idea
I would like to suggest a rewrite of some tools to improve the extensibility and re-usability of the features. I have never worked with UWP before, so I am not very sure about how could dependency injection be configured and used on the view models. I see there are some abstractions injected so i suppose that should be easy to configure for any custom class.

First of all keep into consideration that my suggest would bring some major changes at the view models. Also, at the moment I still have some abstract points without a clear solution.

## My current solution
I will use `JsonYamlToolViewModel` as example:

1. We will need to introduce the concept of 'Services' (or any other way to call it) as the layer where our business code runs. 
2. Move the code usually contained in `TreatQueueAsync` method into the 'services'.
3. It is possible that some changes in the view model could be needed.

For the current example we could use the following interface as the base for any generator-like service.
```
interface IGeneratorService
{
    object Read(string input);
    string Write(object _object);
}
```
Later, we could build other interfaces for different purposes as:
```
interface IIndentedOutput
{
    Indentation Indentation { get; set; }
    # ... here could be any other Indentation related feature
}
interface IConfigurableGenerator : IIndentedOutput # this is wrong, but I will use it as example
{
    void Configure(Action<IConfigurableGenerator> configAction); # here some type capable of collecting the possible configs is needed
}
```
Then we could start building our services once all interfaces are arranged. In example:
```
class JsonConverterService : IGeneratorService, IConfigurableGenerator , etc...
```
For this particular case of generation, the idea is the same used in the current code: transform the string from a defined format into a language object then transform it back to the output format. You could notice that I am not using any `JsonToYaml` like service, so to get this result we could use another service on top of the `Converters`. I will write all the snippet for your consideration:
```
    interface IServiceContainer<T>
    {
        Dictionary<T, IGeneratorService> Services { get; }

        bool TryConvert(string input, out string output, T inputLang, T outputLang);
    }
```
```
    class ServiceTypeAttribute : Attribute
    {
        public int ServiceId { get; set; }
        public ServiceTypeAttribute(int _name)
        {
            ServiceId = _name;
        }
    }
```
```
    class ConverterService : IServiceContainer<GeneratorLanguages>
    {
        private readonly Lazy<Dictionary<GeneratorLanguages, IGeneratorService>> _lazyServices;
        public Dictionary<GeneratorLanguages, IGeneratorService> Services => _lazyServices.Value;

        public ConverterService()
        {
            _lazyServices = new Lazy<Dictionary<GeneratorLanguages, IGeneratorService>>(Initialize);
        }


        public bool TryConvert(string input, out string output, GeneratorLanguages inputLang, GeneratorLanguages outputLang)
        {
            if(!(Services.ContainsKey(inputLang) && Services.ContainsKey(outputLang)) ||
                string.IsNullOrWhiteSpace(input))
            {
                output = string.Empty;
                return false;
            }
            var generic = Services[inputLang].Read(input);
            output = Services[outputLang].Write(generic);
            return true;
        }

        public Dictionary<GeneratorLanguages, IGeneratorService> Initialize()
        {
            var temp = new Dictionary<GeneratorLanguages, IGeneratorService>();
            foreach (var type in Assembly.GetExecutingAssembly().DefinedTypes.Where(t => t.GetCustomAttribute<ServiceTypeAttribute>() is not null && t.GetInterface(nameof(IGeneratorService)) is not null))
            {
                var serviceAttribute = type.GetCustomAttribute<ServiceTypeAttribute>();
                var service = Activator.CreateInstance(type) as IGeneratorService;
                if(Enum.TryParse<GeneratorLanguages>(serviceAttribute.ServiceId.ToString(), out var lang))
                {
                    temp.Add(lang, service);
                }
            }
            return temp;
        }
    }
```
It is not hard to see some holes in this design. It still need to be sharped, as I said at the begging there are some questions without answer. The `Attribute` could be worked out for a cleaner solution. I am using an `IGeneratorService` in the container but how could I configure services without casting? Could this design be generalized somehow to be used with others services (as Encoders)? And I am sure other fissures could be found... Also the classes and interfaces names could be better xD
## Pros 

- Easy incorporation of new serializers/generators/encoders without modifying the current code
- An easy pattern to follow for anyone who decide to add some features
- In general, a more SOLID code

## Conclusions
I will like to do this enhancement if you decide to take it. Before I forget, this would bring changes to the view models. It is possible that some properties/fields needs to be changed. Following with the previous example, these fields should be discarded:
```
        internal const string JsonToYaml = nameof(JsonToYaml);
        internal const string YamlToJson = nameof(YamlToJson);
        private const string TwoSpaceIndentation = "TwoSpaces";
        private const string FourSpaceIndentation = "FourSpaces";
```
Replaced with `GeneratorLanguages` (which I did not included lol) and `Indentation` enums.
Sorry for having to post everything like this here and not through a pull request but I still have not installed VS22, and I am using VS Code to somehow work with this project in net6.0.

youkai95:
I suggested this pattern a few days ago and I decided to give it a try and close #247. This implementation have some details that could be improved and I would like to work on them together.  I hope we could walk to a really SOLID solution that could be compatible with any tool with minor effort.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [x] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

Issue Number: #247

## What is the new behavior?

- A SOLID solution for format converter tool
- Easily extensible to more formats

## Other information

- I would like to begin with an issue that made me expend the latest 30 minutes struggling with it. I have no idea why when the view loads, the format combo boxes does not show the currently selected (or default) format.
- Dont blame me... I dont know where FontIcon glyphs are listed XD
- The way the services are configured does not makes me happy... I would like to share ideas with you to get a sharper design. At the moment it seems ugly to have to put something like `Converters.ConfigureService(value.Value, (service) => service.SetSerializerConfigurations(ConverterConfiguration.Indentation, IndentationMode.Value));` everywhere. In addition to that, we should think about how we could make a better use of the deserializer configurations. Maybe now it would be useless but for future features it could be necessary.

## Quality check


btiteux:
Hello,
Thank you for the contribution, with @veler we need to review these changes as they will modifies lot's of code in the DevToy.
So this pull request will take some times.
Sorry.

youkai95:
> Hello, Thank you for the contribution, with @veler we need to review these changes as they will modifies lot's of code in the DevToy. So this pull request will take some times. Sorry.

I was aware of that, so take the time you need and let me know about any changes the code needs. Also, I would like to point out that this code has an issue in the view that I am not sure how to solve (the first dot in "Other Information")

youkai95:
Few days ago I took a time to review this PR and I decided to change some things. I think I could make it simpler using other pattern...

src/dev/impl/DevToys/Strings/id-ID/JsonYaml.resw
```diff
@@ -144,17 +144,17 @@
   <data name="InvalidYaml" xml:space="preserve">
     <value>YAML yang diinputkan tidak sesuai</value>
   </data>
-  <data name="JsonToYaml" xml:space="preserve">
-    <value>Json ke Yaml</value>
+  <data name="Json" xml:space="preserve">
+    <value>Json</value>
```
Andret2344, line 148: Maybe we should decide now if it should be presented as `Json` or `JSON` in all languages, because now there is no consistency. I vote for `JSON` and `YAML`, as `JSON` is `JavaScript Object Notation` and `YAML` is `Yet Another Markup Language`.

src/dev/impl/DevToys/ViewModels/Tools/Converters/JsonYaml/JsonYamlToolViewModel.cs
```diff
@@ -168,10 +117,27 @@ internal IndentationDisplayPair IndentationMode
...
         /// Gets or sets the input code editor's language.
         /// </summary>
-        internal string? InputValueLanguage
+        internal GeneratorLanguageDisplayPair InputValueLanguage
         {
-            get => _inputValueLanguage;
-            set => SetProperty(ref _inputValueLanguage, value);
+            get
+            {
+                if(Languages is null)
```
Andret2344, line 124: A missing whitespace. Check also other files.

jamescurran:
Is this still under active review/consideration?  I'm also considering a large structural change and it would be best to work on a source code base with the other large structure change already in.

Issue #355, opened 2022-02-08
XML Formatter: Xml declaration always declares encoding as utf-16


jwfxpr:
### Description

As per e633a058, when the XML input omits the declaration, so too does the output. However, when the xml declaration is there, the xml output _always_ declares the encoding to be `utf-16`, regardless of the encoding declared in the input, or if the encoding attribute is omitted from the declaration entirely.

Examples:
![image](https://user-images.githubusercontent.com/20788820/153003674-1bef1e49-637f-49c8-a97a-aa042592a694.png)
![image](https://user-images.githubusercontent.com/20788820/153003749-b6d7bfdf-3990-43f7-b8d7-ad86a8c5e50c.png)

C# internally represents all strings as utf-16, which is probably what causes this.

### Steps To Reproduce

Paste the following example XML in the input field of the XML formatter:

```xml
<?xml version="1.0"?>
<Document version="0.1">
<Element>
<Sub-Element />
</Element>
</Document>
```

![image](https://user-images.githubusercontent.com/20788820/153004367-d4f4e8da-9ff0-431a-a9e4-836935f90117.png)

### Expected behavior

I would not expect the actual content of the input to be changed in any way by the XML formatter, aside from whitespace. Even though _technically_ the output field is in utf-16, if I copy text from a document and then paste that text _back_ to the document after formatting, it will retain the original document's encoding, and the altered declaration will be wrong, leading in many cases to an invalid XML error when that document is used.

In other words, if the `encoding=""` attribute is omitted from the declaration in the input, I expect it to be omitted in the output; if it in present in the input, I expect it to declare the same encoding in the output, _regardless_ of the correctness of the encoding that the output field uses.

### DevToys Version

Version 1.0.2.0 | X64 | RELEASE | b972462 | b972462

### Relevant Assets/Logs

![image](https://user-images.githubusercontent.com/20788820/153003674-1bef1e49-637f-49c8-a97a-aa042592a694.png)
![image](https://user-images.githubusercontent.com/20788820/153003749-b6d7bfdf-3990-43f7-b8d7-ad86a8c5e50c.png)
![image](https://user-images.githubusercontent.com/20788820/153004367-d4f4e8da-9ff0-431a-a9e4-836935f90117.png)

THELEMITED:
So you have to manually get your phone's network translator to support ums, and utm, khtml, umt, mt5, mt6, esp, esp32, Llvm curses #noirerhel

YSternlicht:
[...]

PR #364, opened 2022-02-09
Keep original encoding in declaration

YSternlicht:
## Pull request type

- [ x] Bugfix
- [ ] Feature
- [ ] Build related changes

https://github.com/veler/DevToys/issues/355

Issue Number: N/A

## What is the new behavior?

After having the xml as a string, we replace the hard coded utf-16 to whatever the user input, or removing it if it was omitted
-
-
-

## Quality check

- [ ] Followed the code style guideline as described in [CONTRIBUTING.md](https://github.com/veler/DevToys/blob/main/CONTRIBUTING.md)

### DIFF
--- a/src/dev/impl/DevToys/Helpers/XmlHelper.cs
+++ b/src/dev/impl/DevToys/Helpers/XmlHelper.cs
@@ -3,6 +3,7 @@
 using System;
 using System.IO;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Xml;
 using DevToys.Core;
 using DevToys.Models;
@@ -98,6 +99,18 @@ namespace DevToys.Helpers
                     xmlDocument.Save(xmlWriter);
                 }
 
+                if (xmlDocument.FirstChild.NodeType == XmlNodeType.XmlDeclaration)
+                {
+                    Match match = Regex.Match(xmlDocument.FirstChild.InnerText, @"(?<=encoding\s*=\s*"")[^""]*", RegexOptions.None);
+                    if (match.Success)
+                    {
+                        stringBuilder = stringBuilder.Replace("utf-16", match.Value);
+                    }
+                    else
+                    {
+                        stringBuilder = stringBuilder.Replace("encoding=\"utf-16\"", "");
+                    }
+                }
                 return stringBuilder.ToString();
             }
             catch (XmlException ex)
--- a/src/tests/DevToys.Tests/Helpers/XmlHelperTests.cs
+++ b/src/tests/DevToys.Tests/Helpers/XmlHelperTests.cs
@@ -77,5 +77,18 @@ namespace DevToys.Tests.Helpers
         {
             Assert.AreEqual(expectedResult, XmlHelper.Format(input, Indentation.Minified, newLineOnAttributes));
         }
+
+        [DataTestMethod]
+        [DataRow(null, "")]
+        [DataRow("", "")]
+        [DataRow("<?xml version=\"1.0\" ?><Document version=\"0.1\"><Element /></Document>", "<?xml version=\"1.0\" ?><Document version=\"0.1\"><Element /></Document>")]
+        [DataRow("<?xml version=\"1.0\" encoding=\"utf-8\"?><Document version=\"0.1\"><Element /></Document>", "<?xml version=\"1.0\" encoding=\"utf-8\"?><Document version=\"0.1\"><Element /></Document>")]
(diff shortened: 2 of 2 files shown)


jwfxpr:
Grats on your first PR to the repository @YSternlicht! 😊👍

One suggestion/comment. Here, you check for an XmlDeclaration element:

https://github.com/veler/DevToys/blob/7abeeaf6e9c6100df90cd8f9b9353acb26460e26/src/dev/impl/DevToys/Helpers/XmlHelper.cs#L102

I'd try moving this code block up a few lines to be above the `stringBuilder` declaration, and use pattern matching like this:

```c#
                if (xmlDocument.FirstChild is XmlNodeType.XmlDeclaration declaration)
```

This way, instead of using `stringBuilder.Replace()`, you can simply directly set the value of the [`declaration.Encoding`](https://docs.microsoft.com/en-us/dotnet/api/system.xml.xmldeclaration.encoding?view=net-6.0) property before the XmlDocument is written out, rather than manipulating the output after the fact.

Setting `declaration.Encoding = string.Empty` might remove the element from the declaration, though the documentation does say:

>If an encoding attribute is not included, UTF-8 encoding is assumed when the document is written or saved out.

So that would need a bit of testing. It's not clear if that is literally just the output encoding, or if there would be a `encoding="utf-8"` element instead of no encoding element in the output.

Personally I think that setting `XmlDeclaration.Encoding` is probably more reliable than `Replace()`, since some future refactor might, e.g., change that default encoding somehow. You would end up accidentally replacing the first instance of `"utf-16"` somewhere else in the document. 😊

Also, here:

https://github.com/veler/DevToys/blob/7abeeaf6e9c6100df90cd8f9b9353acb26460e26/src/dev/impl/DevToys/Helpers/XmlHelper.cs#L104

You're scanning the xmlDocument _output_ for the encoding string, rather than the `input`. Is that intended? Either way, if you move the code above the declaration of `stringBuilder`, you'll have to change it over to scanning the `input`:

```c#
                    Match match = Regex.Match(string.Split('>', 2)[0], @"(?<=encoding\s*=\s*"")[^""]*", RegexOptions.None);
```

The `.Split('>', 2)[0]` will make sure that you only match the first element in the input, and don't accidentally match any random `encoding` element somewhere else in the input. There's probably a more elegant way to do that, that's off the top of my head. 😊

YSternlicht:
Thanks @jwfxpr for your comment.

I agree with you that replacing the string after it has been placed in the string builder is not ideal. 
However, a couple of issues forced me to do this.

There are 2 possible encodings to set, one on the xmlWriterSettings, and one on the xmlWriter . Settings property.

https://github.com/YSternlicht/DevToys/blob/7abeeaf6e9c6100df90cd8f9b9353acb26460e26/src/dev/impl/DevToys/Helpers/XmlHelper.cs#L68

https://github.com/YSternlicht/DevToys/blob/7abeeaf6e9c6100df90cd8f9b9353acb26460e26/src/dev/impl/DevToys/Helpers/XmlHelper.cs#L97

jwfxpr:
> I agree with you that replacing the string after it has been placed in the string builder is not ideal. However, a couple of issues forced me to do this.
> ...

Seems like you've already investigated the situation. 😊 I didn't have the opportunity to play with the `XmlDocument` or `XmlWriterSettings` classes before writing the comment, and I think you've covered the same things I would have and I'd have come to the same conclusion as you. 👍

> A possible solution could be to remove the StringBuilder and use a memory stream instead and return that. This will then default to utf-8. That could get very problematic as we will have to specify the encoding before we convert that stream into a string.

Yeah, I agree with you, changing the default to a different default with a more complex process isn't a great solution.

As far as I can see, if we're using `System.Xml` `XmlDocument` and `XmlWriter` to achieve the formatting, this is probably the best workaround available.

@veler, should a couple of simple test cases for this be added to `src/tests/DevToys.Tests/Helpers/XmlHelperTests.cs` in this PR?

veler:
Hi all,

Thank you very much @YSternlicht for opening this PR and proposing this change. 

> As far as I can see, if we're using System.Xml XmlDocument and XmlWriter to achieve the formatting, this is probably the best workaround available.

I agree with this. It seems to be the best workaround so far.

> should a couple of simple test cases for this be added to src/tests/DevToys.Tests/Helpers/XmlHelperTests.cs in this PR?

Definitely! @YSternlicht , would you please be able to add some unit tests? :)

YSternlicht:
Hi @veler 

Sure thing. 

Been a busy week but will try get the tests written over the weekend.

veler:
> Hi @veler 
> 
> Sure thing. 
> 
> Been a busy week but will try get the tests written over the weekend. 

No worries, there's no hurry 😊
Thank you very much!

YSternlicht:
OK.

Tests written.

veler (approved):
Awesome! It looks great! Thank you for this fix! :)

Issue #396, opened 2022-02-15
Checksum Generator won't do comparison


stamminator:
### Description

I can't figure out how to use the checksum compare feature. I would imagine that the intended behavior is for there to be some sort of indicator showing whether a checksum matches or not if the Output Comparer field is populated, but it just seems to be a plain text box with no functionality.

### Steps To Reproduce

1. Drag-and-drop file into Checksum Generator
2. Get the checksum
3. Copy the Output value into Output Comparer
4. Observe that there is no indication of a match or mismatch

### Expected behavior

If the Output Comparer field has anything in it, it should always show a real-time indication of whether it matches the Output field.

### DevToys Version

Version 1.0.2.0 | X64 | RELEASE | b972462 | b972462; OS Version: 19044.1526

### Relevant Assets/Logs

![ApplicationFrameHost_MrcOubYxrq](https://user-images.githubusercontent.com/10633170/154121158-abc27299-6c94-4916-bbaf-1474acf1b614.gif)

PR #406, opened 2022-02-17
Fixing checksum comparison issue (#396)

astegi-dev:
## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Currently, there is no visual indication of the comparison of outputs.

Issue Number: #396 

## What is the new behavior?

When the Output field and the Output Comparsion field both have value, then a text appears stating "The hashes are the same." or "The hashes are different.", whether the two fields have the same or different values.

## Quality check

- [X] Verified that the change work in Release build configuration


veler (changes requested):
Hello,
That's a simple, yet great addition! Thank you for this!

One thing I'd suggest though, the choice of color for the text doesn't work great in Light mode.

image

I'd suggest using an InfoBar, like in Number Base Converter.

image

See here: 
https://github.com/veler/DevToys/blob/5e423b37edee81af1e83f6a202d8073036658c8c/src/dev/impl/DevToys/Views/Tools/Converters/NumberBaseConverter/NumberBaseConverterToolPage.xaml#L31-L34

You can also use the `IsClosable` property of this control to hide the close button at the right (I think it would make sense to not being able to close this information).

Even better, you could make the info bar yellow or red when the hash isn't matching, and green (Severity = Success) when it matches. What do you think about it?

astegi-dev:
Yeah, sure, I'm fixing that.

I am not an experienced UWP developer, I didn't know such control exists, but definitely makes more sense. :)

astegi-dev:
Screenshots of the modification:
When hashes match:
image

When hashes does not match:
image

On light background:
image
image

veler (approved):
Hello,
Sorry for the late answer. It looks all good now :D Thank you so much for this contribution! Love it !

PR #430, opened 2022-02-27
Format from/to custom number base

### LINKED ISSUE #384: Number Base Converter - Custom base
### What's the Problem?

_No response_

### Solution/Idea

Now number base formatter allows only converting between binary, octal, decimal and hexadecimal number bases. But It is possible to use different bases, e.g. 7, 3 or 12. How about allowing to decide about input and output number base for conversion by typing number between 1 (unary system) and (let's say) 36 (all uppercase alphanumerics).

### Alternatives

Using [this](https://www.rapidtables.com/convert/number/base-converter.html) website

### Priorities

| Capability | Priority |
| :---------- | :------- |
| Allow to choose input and output number base | Must |
| Properly handle negative numbers | Should |
| Have no upper limits of numbers | Could |
| Allow to add "another" row with different (or same...) number base output  | Could |

### DevToys Version

Version 1.0.3.0 | X64 | RELEASE | 7c7c02a | 7c7c02a

### Comments

_No response_

youkai95:
Ok, now I think it has all the required functionalities. Here I left some features available like allowing formatting even for those base numbers using custom dictionaries. That idea could be improved as it would be nice to allow user to change the default character and group size... Yes, it is hardcoded at the moment.

I am not sure at all if I faced this problem in a correct way. If you have any comment regarding my code, please share it with me to improve it.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: resolve #384

## What is the new behavior?

- Convert numbers from/to different bases
- Provide and advanced view to allow complex functionalities.
- Allows formatting for custom dictionaries (need work)

## Quality check


src/dev/impl/DevToys/Views/Tools/Converters/NumberBaseConverter/NumberBaseConverterToolPage.xaml.cs
```diff
@@ -47,21 +47,22 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
...
+            //    }
+            //    else if (NumberBaseHelper.IsValidHexadecimal(clipBoardContent!))
+            //    {
+            //        ViewModel.InputBaseNumber = NumberBaseFormat.Hexadecimal;
+            //    }
 
-                ViewModel.InputValue = parameters.ClipBoardContent;
-            }
+            //    ViewModel.InputValue = parameters.ClipBoardContent;
+            //}
```
youkai95, line 65: It would be good to revert this changes and keep this feature. This could require big changes but if you give me any idea I will try to fix it.

youkai95, line 65: It would be good to revert this changes and keep this feature. This could require big changes but if you give me any idea I will try to fix it.

veler:
> @veler @btiteux this PR is ready to review 😁 There are few details that would be nice to improve, but overall, it is ready to check.

Hello :)
Sorry for the very late answer. I'm still very busy lately. I will take a look now. :)

src/dev/impl/DevToys/Views/Tools/Converters/NumberBaseConverter/NumberBaseConverterToolPage.xaml
```diff
@@ -26,14 +32,30 @@
                     Style="{StaticResource RightAlignedToggleSwitchStyle}"
                     IsOn="{x:Bind ViewModel.IsFormatted, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}"/>
             </controls:ExpandableSettingControl>
+            <controls:ExpandableSettingControl
+                Title="{x:Bind ViewModel.Strings.AdvancedModeLabel}">
+                <controls:ExpandableSettingControl.Icon>
+                    <FontIcon Glyph="&#xF7B2;" />
```
veler, line 38: Try `&#xF1EE;`, please.

image

src/dev/impl/DevToys/Views/Tools/Converters/NumberBaseConverter/NumberBaseConverterToolPage.xaml
```diff
@@ -26,14 +32,30 @@
...
 
-        <controls:CustomTextBox
+        <Grid Visibility="{x:Bind ViewModel.AdvancedMode, Mode=OneWay, Converter={StaticResource InvertedBooleanToVisibilityConverter}}">
+            <nbc:BasicNumberBaseConverterControl/>
+        </Grid>
+        <Grid Visibility="{x:Bind ViewModel.AdvancedMode, Mode=OneWay, Converter={StaticResource BooleanToVisibilityConverter}}">
+            <nbc:AdvancedNumberBaseConverterControl/>
+        </Grid>
+
+        <!--<controls:CustomTextBox
```
veler, line 58: You can remove this commented code.

src/dev/impl/DevToys/Views/Tools/Converters/NumberBaseConverter/NumberBaseConverterToolPage.xaml
```diff
@@ -26,14 +32,30 @@
...
         </StackPanel>
 
         <muxc:InfoBar 
             IsOpen="{x:Bind ViewModel.IsInfoBarOpen, Mode=OneWay}" 
             Severity="Warning" 
             Message="{x:Bind ViewModel.InfoBarMessage, Mode=OneWay}" />
 
-        <controls:CustomTextBox
+        <Grid Visibility="{x:Bind ViewModel.AdvancedMode, Mode=OneWay, Converter={StaticResource InvertedBooleanToVisibilityConverter}}">
+            <nbc:BasicNumberBaseConverterControl/>
```
veler, line 52: No need to place this control in a Grid. Just place the Visibility binding directly on the control.

src/dev/impl/DevToys/Views/Tools/Converters/NumberBaseConverter/BasicNumberBaseConverterControl.xaml
```diff
@@ -0,0 +1,43 @@
...
+    xmlns:DevToys="using:DevToys"
+    xmlns:converters="using:DevToys.UI.Converters"
+    mc:Ignorable="d"
+    d:DesignHeight="400"
+    d:DesignWidth="600">
+    <UserControl.Resources>
+        <DevToys:LanguageManager x:Key="LanguageManager"/>
+    </UserControl.Resources>
+    
+    <StackPanel>
```
veler, line 17: Add `Spacing="12"`. Otherwise the spacing isn't good:

image

src/dev/impl/DevToys/Views/Tools/Converters/NumberBaseConverter/AdvancedNumberBaseConverterControl.xaml
```diff
@@ -0,0 +1,73 @@
...
+    mc:Ignorable="d"
+    d:DesignHeight="400"
+    d:DesignWidth="600">
+
+    <UserControl.Resources>
+        <converters:BooleanToVisibilityConverter x:Key="BooleanToVisibilityConverter"/>
+    </UserControl.Resources>
+    
+    <StackPanel Margin="0,0,0,16" Spacing="12">
+        <StackPanel.ChildrenTransitions>
```
veler, line 19: No need of animation here in my opinion.

src/dev/impl/DevToys/Views/Tools/Converters/NumberBaseConverter/AdvancedNumberBaseConverterControl.xaml
```diff
@@ -0,0 +1,73 @@
...
+                    SelectedItem="{x:Bind ViewModel.InputNumberFormat, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}"
+                    ItemsSource="{x:Bind ViewModel.BaseNumbers}">
+                </ComboBox>
+            </controls:ExpandableSettingControl>
+            
+            <TextBox Visibility="{x:Bind ViewModel.UseInputCustomDictionary, Mode=OneWay, Converter={StaticResource BooleanToVisibilityConverter}}" 
+                     Text="{x:Bind ViewModel.InputCustomDictionary, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}"/>
+            
+            <controls:CustomTextBox
+                Header="{x:Bind ViewModel.Strings.DecimalLabel}"
```
veler, line 43: Interestingly, the header doesn't show up on my side:

image

veler:
Custom base doesn't seem to work.

image


I'm also facing some error banner when enabling/disabling `Format number`. 
Also sometimes I'd type something in the Basic mode input and would get an error about the fact that the Custom Base dictionary is wrong. You should really separate the view model of the Basic and Advanced mode so they don't conflict.

src/dev/impl/DevToys/Views/Tools/Converters/NumberBaseConverter/AdvancedNumberBaseConverterControl.xaml
```diff
@@ -0,0 +1,73 @@
...
+            </TransitionCollection>
+        </StackPanel.ChildrenTransitions>
+
+        <StackPanel Spacing="4" >
+            <controls:ExpandableSettingControl
+                x:Name="InputDictionaryValue"
+                Grid.Row="2"
+                Title="{x:Bind ViewModel.Strings.InputDictionaryLabel}">
+                <controls:ExpandableSettingControl.Icon>
+                    <FontIcon />
```
veler, line 31: Try the glyph `&#x0065;`

image

src/dev/impl/DevToys/Views/Tools/Converters/NumberBaseConverter/AdvancedNumberBaseConverterControl.xaml
```diff
@@ -0,0 +1,73 @@
...
+                Text="{x:Bind ViewModel.InputValue, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}"/>
+        </StackPanel>
+
+        <StackPanel Spacing="4" >
+            <controls:ExpandableSettingControl
+                x:Name="OutputDictionaryValue"
+                Grid.Row="2"
+                Title="{x:Bind ViewModel.Strings.OutputDictionaryLabel}">
+                <controls:ExpandableSettingControl.Icon>
+                    <FontIcon />
```
veler, line 54: Try `&#x0064;` here :)

image

src/dev/impl/DevToys/ViewModels/Tools/Converters/NumberBaseConverter/NumberBaseFormatter.cs
```diff
@@ -11,6 +13,52 @@ internal static class NumberBaseFormatter
...
         private static NumberBaseConverterStrings Strings => LanguageManager.Instance.NumberBaseConverter;
 
+        /// <summary>
+        /// Based on <see cref="System.ParseNumbers"/>
+        /// https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/ParseNumbers.cs
+        /// </summary>
+        /// <param name="value">Current value</param>
+        /// <param name="baseNumber">Current base number <see cref="NumberBaseFormat"/></param>
+        /// <returns>Value converted to <see cref="NumberBaseFormat.Decimal"/></returns>
+        //public static long? StringToBase(string value, NumberBaseFormat baseNumber)
```
veler, line 23: Please delete this commented code if you don't use it.

src/dev/impl/DevToys/Strings/en-US/NumberBaseConverter.resw
```diff
@@ -179,4 +179,13 @@
   <data name="SearchKeywords" xml:space="preserve">
     <value>Binary Octal Decimal Hexadecimal</value>
   </data>
+  <data name="AdvancedModeLabel" xml:space="preserve">
```
veler, line 182: Please apply these changes to every other language (without translating it). This will help translators to know what is to be translated.

src/dev/impl/DevToys/Views/Tools/Converters/NumberBaseConverter/NumberBaseConverterToolPage.xaml.cs
veler: What's the challenge you're encounter here? Is the problem to determine whether it's an advanced mode or not?

src/dev/impl/DevToys/ViewModels/Tools/Converters/NumberBaseConverter/BasicNumberBaseConverterControlViewModel.cs
```diff
@@ -0,0 +1,291 @@
...
+using DevToys.UI.Controls;
+using Microsoft.Toolkit.Mvvm.ComponentModel;
+using Microsoft.Toolkit.Mvvm.Messaging;
+using Windows.UI.Xaml;
+using DevToys.Core;
+using DevToys.Messages;
+
+namespace DevToys.ViewModels.Tools.Converters.NumberBaseConverter
+{
+    [Export(typeof(BasicNumberBaseConverterControlViewModel))]
```
veler, line 25: No need to do an Export since you instantiate it manually in the dependency property in .xaml.cs

src/dev/impl/DevToys/Views/Tools/Converters/NumberBaseConverter/AdvancedNumberBaseConverterControl.xaml.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
```
veler, line 1: Please add `#nullable enabled`

src/dev/impl/DevToys/ViewModels/Tools/Converters/NumberBaseConverter/NumberBaseFormatFactory.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Threading.Tasks;
+using DevToys.Models;
+
+namespace DevToys.ViewModels.Tools.Converters.NumberBaseConverter
+{
+    internal class NumberBaseFormatBuilder
+    {
+        public string DisplayName { get; set; } = string.Empty;
+
+        //public Radix Value { get; set; } = Radix.Custom;
```
veler, line 15: please remove it if it's not useful.

src/dev/impl/DevToys/Models/NumberBaseDictionary.cs
```diff
@@ -0,0 +1,68 @@
+﻿using System;
```
veler, line 1: Please add `#nullable enabled`

src/dev/impl/DevToys/Models/NumberBaseDictionary.cs
```diff
@@ -0,0 +1,68 @@
...
+        public static readonly NumberBaseDictionary RFC4648Base64Dictionary = new(
+            _dictionary: "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/".ToCharArray(),
+            formatting: false);
+
+        public static readonly NumberBaseDictionary RFC4648Base64UrlEncodeDictionary = new(
+             _dictionary: "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_".ToCharArray(),
+             formatting: false);
+        public static NumberBaseDictionary DefaultDictionary { get; } = RFC4648Base64Dictionary;
+
+        public char[] Dictionary { get; }
```
veler, line 32: I'd suggest using a `string` instead. You can still do `return Dictionary[index];` on line 50. Using a string would avoid us to allocate a string and then a char array right after.

src/dev/impl/DevToys/Messages/ChangeNumberFormattingMessage.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
```
veler, line 1: Here too

src/dev/impl/DevToys/Messages/ChangeInfoBarStatusMessage.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
```
veler, line 1: `#nullable enabled`

veler (changes requested):
Quite a big PR, good job :)
I encountered some bugs though, or didn't understand how to use it (in particular the custom base).

I'm also hitting some bugs that seems to come from the Advanced mode when I'm editing something in Basic mode (for example, the Format number option).

Thank you :)

src/dev/impl/DevToys/Views/Tools/Converters/NumberBaseConverter/NumberBaseConverterToolPage.xaml.cs
youkai95: Yes, the tool opens through NumberBaseConverter main view but the functionalities are inside Basic/Advanced view models, who are not declared inside the main view. However, I just realized that I could get this done using messages, isn't it? Still remains the problem of selecting between basic or advanced view...

src/dev/impl/DevToys/Views/Tools/Converters/NumberBaseConverter/AdvancedNumberBaseConverterControl.xaml
youkai95: Yes, that's really interesting. I haven't noticed it has a wrong label as I never saw it when testing the app. This issue also happens in the output textbox...

src/dev/impl/DevToys/ViewModels/Tools/Converters/NumberBaseConverter/BasicNumberBaseConverterControlViewModel.cs
youkai95: It is there any way to prevent manual instantiation for both view models?

src/dev/impl/DevToys/Views/Tools/Converters/NumberBaseConverter/NumberBaseConverterToolPage.xaml.cs
veler: Ah I see. You should be able to do this by:
1. Set `x:Name="AdvancedNumberBaseConverterControl"` on the advanced control in the main view
2. And then, something like:

```
ViewModel.AdvancedMode = true;
AdvancedNumberBaseConverterControl.ViewModel.Property = something;
```

youkai95:
> Custom base doesn't seem to work, or I don't understand how it should work.
> 
> image
> 

If you use a custom dictionary for input, then you should write an input value using that dictionary. Using you example, a correct input should be a value with characters:

```
 1234abcd
(01234567)
```
The line below correspond to the values that every character has when converting to decimal base.

> I'm also facing some error banner when enabling/disabling `Format number`. Also sometimes I'd type something in the Basic mode input and would get an error about the fact that the Custom Base dictionary is wrong. You should really separate the view model of the Basic and Advanced mode so they don't conflict.

I am not sure about how to replicate this error. I would be grateful if you could provide me an example to simulate it. I have decided to keep the info box into the parent component to avoid having duplicated code as it is the same behavior for both views.

src/dev/impl/DevToys/ViewModels/Tools/Converters/NumberBaseConverter/BasicNumberBaseConverterControlViewModel.cs
veler: Yes, alternatively, you could import this view model into the main one (NumberBaseConverterToolViewModel) with something like:

```
[ImportingConstructor]
public NumberBaseConverterToolViewModel(BasicNumberBaseConverterControlViewModel viewmodel)
{
}
```

youkai95:
@veler I already pushed all the reviews you asked for. I also found some bugs and misbehavior in the code that should be solved now.

veler:
> > Custom base doesn't seem to work, or I don't understand how it should work.
> > image
> 
> If you use a custom dictionary for input, then you should write an input value using that dictionary. Using you example, a correct input should be a value with characters:
> 
> ```
>  1234abcd
> (01234567)
> ```
> 
> The line below correspond to the values that every character has when converting to decimal base.
> 
> > I'm also facing some error banner when enabling/disabling `Format number`. Also sometimes I'd type something in the Basic mode input and would get an error about the fact that the Custom Base dictionary is wrong. You should really separate the view model of the Basic and Advanced mode so they don't conflict.
> 
> I am not sure about how to replicate this error. I would be grateful if you could provide me an example to simulate it. I have decided to keep the info box into the parent component to avoid having duplicated code as it is the same behavior for both views.

I see! Thanks for the explanation! It wasn't clear to me by looking at the UI that works this way. Should we consider clarifying it in the UI? What do yout think?

> @veler the reviews you asked for should be solved with this push, sorry for the late. I also found some bugs and misbehavior in the code that should be solved now.

Thank you for this too. I gave it a try and it seems better :D
I added a few other comments to the code. Can you please address them when you have some time?
There's also a merge conflict to address.

After all this, I think it will be good to merge.

Thanks again!

youkai95:
> Should we consider clarifying it in the UI? What do yout think?

I think that would be a good idea. If you were aware of this tool capabilities and you got confused, what could we expect from others? XD Also, I will try to discover what is happening with the label that doesnt get shown.

>I added a few other comments to the code. Can you please address them when you have some time?
There's also a merge conflict to address.

Ok, I could address them in few time and do the merge. I also noticed that I miss some old changes...

src/dev/impl/DevToys/Views/Tools/Converters/NumberBaseConverter/AdvancedNumberBaseConverterControl.xaml
youkai95: It seems like when the UserControl (`AdvancedNumberBaseConverterControl` in this case) is not visible the values for `Header` in `CustomTextBox` are not correctly bound, so it is set to null. I cannot tell why is this happening.

veler (approved):
Hello,
Sorry for the delay.
It looks great  now ! :D I will merge it. Thank you for the contribution, much appreciated! :)

Issue #448, opened 2022-03-13
Cut does not work properly


morning4coffe-dev:
### Description

When I was testing the localization update I discovered that the Cut feature in the text box does not work properly.

### Steps To Reproduce

1. Go to any tool with the text box.
2. Write some text to it.
3. Select some text.
4. Right-click it.
5. Hit Cut (or Ctrl+X)
6. Does not work.

### Expected behavior

The selected text should get removed and it should be added to the clipboard.

### DevToys Version

1.0.4.0 | X64 | RELEASE | OS: 22572.1

### Relevant Assets/Logs

![image](https://user-images.githubusercontent.com/78549750/158082442-402c2d1c-b134-47ac-b961-8c891a8bfc84.png)

niyari:
Would adding it here fix it?

https://github.com/veler/DevToys/blob/232b783ef245c39ed8a16c1bf401e41221a38225/src/dev/impl/DevToys/UI/Controls/CustomTextBox.xaml.cs#L625

>         private void TextBox_CuttingToClipboard(TextBox sender, TextControlCuttingToClipboardEventArgs args)
>         {
>             CopyTextBoxSelectionToClipboard();
>             sender.SelectedText = string.Empty;           
>             args.Handled = true; 
>         }
> 

https://github.com/veler/DevToys/blob/232b783ef245c39ed8a16c1bf401e41221a38225/src/dev/impl/DevToys/UI/Controls/CustomTextBox.xaml.cs#L642

>        private void RichEditBox_CuttingToClipboard(RichEditBox sender, TextControlCuttingToClipboardEventArgs args)
>         {
>             CopyRichEditBoxSelectionToClipboard();
>             sender.TextDocument.Selection.Cut(); // There has to be a better way.
>             args.Handled = true;
>         }
>

PR #458, opened 2022-03-20
Fix textbox cut command

niyari:
## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #448 

## What is the new behavior?

- Disabled cut command for read-only textbox.
- Added post-processing for text cut command.

## Other information

Disabled cut command for read-only textbox.
This change eliminates access denied errors.
image

> Feature name: Failed to cut from custom text box
> Custom message: 
> Exception message: Access is denied. (Exception from HRESULT: 0x80070005)
> Exception stack trace:
>    at System.Runtime.InteropServices.McgMarshal.ThrowOnExternalCallFailed(Int32, RuntimeTypeHandle) + 0x21
>    at __Interop.ComCallHelpers.Call(__ComObject, RuntimeTypeHandle, Int32) + 0xb8
>    at __Interop.ForwardComStubs.Stub_17[TThis](__ComObject, Int32) + 0x24
>    at Windows.UI.Text.ITextRange__Impl.Dispatcher.global::Windows.UI.Text.ITextRange.Cut() + 0x18
>    at DevToys.UI.Controls.CustomTextBox.ExecuteCutCommand() + 0x3b

## Quality check

### DIFF
--- a/src/dev/impl/DevToys/UI/Controls/CustomTextBox.xaml.cs
+++ b/src/dev/impl/DevToys/UI/Controls/CustomTextBox.xaml.cs
@@ -139,7 +139,11 @@ namespace DevToys.UI.Controls
 
         private bool CanExecuteCutCommand()
         {
-            return RichEditBox != null && RichEditBox.TextDocument.Selection.Length != 0 && IsEnabled && RichEditBox.TextDocument.CanCopy();
+            return RichEditBox != null
+                   && RichEditBox.TextDocument.Selection.Length != 0
+                   && IsEnabled
+                   && RichEditBox.TextDocument.CanCopy()
+                   && IsReadOnly == false;
         }
 
         private void ExecuteCutCommand()
--- a/src/dev/impl/DevToys/UI/Controls/CustomTextBox.xaml.cs
+++ b/src/dev/impl/DevToys/UI/Controls/CustomTextBox.xaml.cs
@@ -140,10 +140,10 @@ namespace DevToys.UI.Controls
         private bool CanExecuteCutCommand()
         {
             return RichEditBox != null
+                   && !IsReadOnly
                    && RichEditBox.TextDocument.Selection.Length != 0
                    && IsEnabled
-                   && RichEditBox.TextDocument.CanCopy()
-                   && IsReadOnly == false;
+                   && RichEditBox.TextDocument.CanCopy();
         }
 
         private void ExecuteCutCommand()
@@ -207,7 +207,7 @@ namespace DevToys.UI.Controls
 
         private bool CanExecuteDeleteCommand()
         {
-            return RichEditBox != null && RichEditBox.TextDocument.Selection.Length != 0 && IsEnabled && !IsReadOnly;
+            return RichEditBox != null && !IsReadOnly && RichEditBox.TextDocument.Selection.Length != 0 && IsEnabled;
         }
 
         private void ExecuteDeleteCommand()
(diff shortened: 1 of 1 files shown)


src/dev/impl/DevToys/UI/Controls/CustomTextBox.xaml.cs
```diff
@@ -139,7 +139,11 @@ public CustomTextBox()
 
         private bool CanExecuteCutCommand()
         {
-            return RichEditBox != null && RichEditBox.TextDocument.Selection.Length != 0 && IsEnabled && RichEditBox.TextDocument.CanCopy();
+            return RichEditBox != null
```
niyari, line 142: The one line became longer, so it was made into multiple lines. It can be undone with code style.

src/dev/impl/DevToys/UI/Controls/CustomTextBox.xaml.cs
```diff
@@ -139,7 +139,11 @@ public CustomTextBox()
...
+                   && RichEditBox.TextDocument.Selection.Length != 0
+                   && IsEnabled
+                   && RichEditBox.TextDocument.CanCopy()
+                   && IsReadOnly == false;
```
niyari, line 146: Do I need to shorten it to fit other code?
> && !IsReadOnly;

youkai95: There is no need to use an explicit comparison, it would be better to use `!IsReadOnly`.

Also, about long conditionals, I would recommend you to order the conditions based on what could be more likely to happen to break the condition. This is just a small improvement but it would be nice to stop the conditional early and avoid extra (or even heavy!) work ;)

niyari: > This is just a small improvement but it would be nice to stop the conditional early and avoid extra (or even heavy!) work

That is correct.

```
       private bool CanExecuteCutCommand()
        {
            Logger.Log("RichEditBox", $"IsEnabled {IsEnabled} / CanCopy {RichEditBox.TextDocument.CanCopy()} / IsReadOnly {IsReadOnly}");
           ...

```
I added a small code to confirm, logged it, and searched for other code that would set the change of state.
RichEditBox.IsEnabled seems to always be True; I could not find any code to change it to False. (However, I will not change it now.)
In the case of CanExecuteCutCommand ( and CanExecuteDeleteCommand), IsReadOnly is evaluated early.

```
        private bool CanExecuteCutCommand()
        {
            return RichEditBox != null
                   && !IsReadOnly
                   && RichEditBox.TextDocument.Selection.Length != 0
                   && IsEnabled
                   && RichEditBox.TextDocument.CanCopy();
        }
```

youkai95: >I added a small code to confirm, logged it, and searched for other code that would set the change of state.
RichEditBox.IsEnabled seems to always be True; I could not find any code to change it to False. (However, I will not change it now.)

I am not sure if this tool is used in an environment where it is disabled, however that condition is not wrong. I could recommend the following design as a nice one for a general use:

```
            return IsEnabled
                   && !IsReadOnly
                   && RichEditBox != null
                   && RichEditBox.TextDocument.Selection.Length != 0
                   && RichEditBox.TextDocument.CanCopy();
```

The first two comparisons are simple boolean checks and the following statements depends on if those two are true. Anyways I dont consider that your proposal should be necessarily changed as any of those statements are simpler enough to be a problem.

veler (approved):
Hello,
Thank you very much for finding and fixing this bug. It looks all good :)

Issue #473, opened 2022-03-28
Unix Timestamp Converter - Timezone select


Andret2344:
### Solution/Idea

When using timestamp converter, the date and time fields shows correct result, but doesn't allow me to change the time zone. It only takes the UTC (one of timezones) and my current (CEST). 
![image](https://user-images.githubusercontent.com/40674613/160441269-43efbb60-14b4-4c3d-8aae-baf9f6c2b400.png)

### Alternatives

I must change my PC timezone.

### Priorities

| Capability | Priority |
| :---------- | :------- |
| Allow to choose timezone | Must |
| Merge two existing datetime rows into one with timezone select | Should |
| Have a possibility to add another row with another timezone | Could |
| Force me to use my current timezone (maybe as a default only) | Won't |

### DevToys Version

Version 1.0.5.0 | X64 | RELEASE | 99757dc | 99757dc

niyari:
(WIP)
![image](https://user-images.githubusercontent.com/2970321/161846204-bcca1549-1dd1-480a-8531-38ea10fae034.png)

Andret2344:
Looks really great! I have just a few suggestions:
* What's the point of leaving UTC, if it's one of the timezones?
* How about adding (where exist) named timezones in curly brackets (I.e. GMT, CET, UTC, PST etc.)? It'd look like `+01:00 (CET)` and `+02:00 (CEST)`.

niyari:
> What's the point of leaving UTC, if it's one of timezones?

As I was updating this tool, I realized that the feature users want is "time conversion between two time zones."
For example, I have a scenario where I receive an email about a domain (e.g. .io ) and I want to compare Japan (+0900 JST) with the British Indian Ocean Territory (+0600).

![image](https://user-images.githubusercontent.com/2970321/161914756-b04d99cd-4a99-4843-a699-6577db1c6a61.png)
[TimeZonesMap - Wikipedia](https://upload.wikimedia.org/wikipedia/commons/8/88/World_Time_Zones_Map.png)

I think a better tool would be to set `Zone 1` to UTC and `Zone 2` to LocalTime as default values, and allow users to select any time zone.

> How about adding (where exist) named timezones in curly brackets (I.e. GMT, CET, UTC, PST etc.)? It'd look like +01:00 (CET) and +02:00 (CEST).

![image](https://user-images.githubusercontent.com/2970321/161916040-de7cdab3-350e-4333-8ab0-34b86ed3baf8.png)
[...]

PR #509, opened 2022-04-30
Updated Unix Timestamp Converter

niyari:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #473

## What is the new behavior?

image

- Time zone support.
- Support for time carry forward and backward.
- Resolved performance degradation that occurs when values are out of range.

Support for time carry forward and backward.

image

## Other information

- ~~Conflict with the addition of a `Now` button needs to be resolved.~~

## Quality check


niyari:
Looks good to display in text, InfoBar can be removed.
Timestamp to .NET Ticks (If features are to be added here.) #512 
image

veler:
Hello :)
Thank you for getting the PR up, and sorry for the late answer.

A few remarks on the screenshot you shared:

The following info is great, but they visually don't align with the rest of the UI on the left:

image

**Suggestion:**

What about putting this info either below the `Time Zone` or `Year Month Day...` input?
Additionally, perhaps it could be place into a Grid with a lighter background color, like through this style:

https://github.com/veler/DevToys/blob/b11f267ef738501bbf993c677013777cb764b1aa/src/dev/impl/DevToys/Views/Tools/Graphic/ColorBlindnessSimulator/ImagePreview.xaml#L12-L23

Which looks like this:

image

Or even, an expandable container, like in the setting page:

image

Here is how it's made: 
https://github.com/veler/DevToys/blob/da659990b1185011edb8182ad340e37e6b5f6611/src/dev/impl/DevToys/Views/Tools/Settings/SettingsToolPage.xaml#L142-L165

What do you think? :)

Thank you !

niyari:
`Timezone info` UI is derived from the `Text Case Converter`.
Of the suggestions for unifying visual information, we have 3 candidates.
`expandable container` hides much needed information for users living in areas that do not use daylight savings time.

image

In the case of A and B, B is usually more compact on the screen.
However, DevToys is expected to be used primarily with some kind of work.
In order for DevToys to be a viable tool with the smallest screen size, all the operable items need to be gathered at the top.

image


It would be best if the screen layout could be changed according to the window size like on the web, but first we would like to display it at the bottom of the C plan.
image

veler:
Hello,

Thanks for your quick answer :)

I like A plan better!
Although you are right, either plan A, B or C, there is an issue with Compact Mode.
Good thing is, by using VisualState, we can actually change the layout on Compact Mode. We do something like this for TextDiff tool:
https://github.com/veler/DevToys/blob/b11f267ef738501bbf993c677013777cb764b1aa/src/dev/impl/DevToys/Views/Tools/Text/TextDiff/TextDiffToolPage.xaml#L14-L30

We hide the input fields and for this, we have to change a bit the layout so it resize accordingly. Perhaps we can do the same here so some information displayed horizontally turn to be displayed vertically?

niyari:
Responsive (changes with window width) is going to be better regardless of mode.
I would like to make this a separate PR, what do you think?

Furthermore, I found that after changing to compact mode, changing the window size causes the application to stop.
I will submit a new issue.

veler:
Hello,
Thanks for your answer. I'm fine with addressing that in another PR, perhaps I can even do that myself later.
There are some merge conflict in this PR apparently :)

niyari:
How about this?(use AdaptiveTrigger)
image

veler:
> How about this?(use AdaptiveTrigger) image

I like that better! In particular the year/month/day going on a second line. 
I find a bit more weird, however, that the time zone information is jumping from above year/month/day to below year/month/day.

niyari:
I think this will work.

veler:
Can we keep the TimeZone information just below the TimeZone combo box?

niyari:
In commit c3027e0, the Timezone combo box is back to fixed.
image

src/dev/impl/DevToys/Views/Tools/Converters/Timestamp/TimestampToolPage.xaml
```diff
@@ -9,41 +9,206 @@
...
+        <StackPanel
+            Margin="0,12,0,0">
             <muxc:InfoBar
                 Severity="Warning"
                 Message="{x:Bind ViewModel.Strings.InvalidValue}"
                 IsOpen="{x:Bind ViewModel.IsInputInvalid, Mode=OneWay}"
-                IsClosable="False"/>
+                IsClosable="False" />
+
+            <TextBlock
```
veler, line 61: I have trouble understanding the goal of this `Current date and time` label. It's just above `Timezone` and it doesn't show constantly the current time in the UI.

src/dev/impl/DevToys/Views/Tools/Converters/Timestamp/TimestampToolPage.xaml
```diff
@@ -9,41 +9,206 @@
...
+                    IsStaggeringEnabled="True"
+                    FromVerticalOffset="50" />
             </TransitionCollection>
         </StackPanel.ChildrenTransitions>
 
-        <StackPanel Margin="0,12,0,0">
-
+        <StackPanel
+            Margin="0,12,0,0">
             <muxc:InfoBar
```
veler, line 55: Silly question: does this info bar still get displayed at some point? I tried to play with all the fields to "break" it, but at the end it seems like the timestamp is being fixed by itself, which makes that I don't succeed to get this info bar to be displayed, which makes me think that perhaps we can remove it?

src/dev/impl/DevToys/Views/Tools/Converters/Timestamp/TimestampToolPage.xaml
```diff
@@ -9,41 +9,206 @@
...
+                ItemsSource="{x:Bind ViewModel.TimeZoneDisplayNameCollection}" />
+
+        </StackPanel>
+        <!-- timezone / -->
+
+        <!-- Timezone info   -->
+        <StackPanel
+            x:Name="Timezoneinfo">
+            <UserControl>
+                <UserControl.Resources>
```
veler, line 91: Why using a UserControl here? That CardStyle can be placed in a Page.Resources tag, after line 11.

src/dev/impl/DevToys/Views/Tools/Converters/Timestamp/TimestampToolPage.xaml.cs
```diff
@@ -52,11 +52,12 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
             {
                 if (long.TryParse(parameters.ClipBoardContent, out long timestamp))
                 {
-                    ViewModel.Timestamp = timestamp;
+                    ViewModel.CurrentTimestamp = timestamp;
                 }
                 else if (DateTime.TryParse(parameters.ClipBoardContent, out DateTime localDateTime))
                 {
-                    ViewModel.Timestamp = new DateTimeOffset(localDateTime.ToUniversalTime()).ToUnixTimeSeconds();
+                    // TODO:
```
veler, line 59: To do what? :)

src/dev/impl/DevToys/ViewModels/Tools/Converters/Timestamp/TimestampToolViewModel.cs
```diff
@@ -14,9 +16,22 @@ namespace DevToys.ViewModels.Tools.Timestamp
     public sealed class TimestampToolViewModel : ObservableRecipient, IToolViewModel
     {
         private bool _isInputInvalid;
-        private double _timestamp;
-        private DateTime _utcDateTime;
-        private Lazy<DateTime> _localDateTime = null!;
+
+        internal List<string> TimeZoneDisplayNameCollection = TimestampToolHelper.ZoneInfo.DisplayNames;
```
veler, line 20: We can make this `IReadOnlyList<string>`

src/dev/impl/DevToys/ViewModels/Tools/Converters/Timestamp/TimestampToolViewModel.cs
```diff
@@ -14,9 +16,22 @@ namespace DevToys.ViewModels.Tools.Timestamp
...
+        private readonly Dictionary<string, string> _timeZoneCollection = TimestampToolHelper.ZoneInfo.TimeZones;
```
veler, line 21: `IReadOnlyDictionary<string, string>`

src/dev/impl/DevToys/ViewModels/Tools/Converters/Timestamp/TimestampToolViewModel.cs
```diff
@@ -28,275 +43,193 @@ internal bool IsInputInvalid
             set => SetProperty(ref _isInputInvalid, value);
         }
 
-        internal double Timestamp
+        internal string DSTInfoText
```
veler, line 46: For future maintainers (and for me 😅), it would be nice to add a description at the top of every properties in this document. You can do it with the following syntax:

```
        /// <summary>
        /// Gets or sets the daylight saving time local time displayed in the UI (or whatever describes better this property 😉)
        /// </summary>
        internal string DSTInfoLocalDateTime
        {
            get => _dstInfoLocalDateTime;
            set => SetProperty(ref _dstInfoLocalDateTime, value);
        }
```

src/dev/impl/DevToys/Helpers/TimestampToolHelper.cs
```diff
@@ -0,0 +1,108 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using DevToys.Core;
+using DevToys.Models;
```
veler, line 11: It looks like several of these `using` statements are useless. Pro tip: in Visual Studio, you can use `Ctrl+R, Ctrl+G` to re-order and clear the `using` statements automatically :)

veler (changes requested):
I looked at the PR, built the branch and played with it. It looks very good overall 😁 Good job! Just a few details and here and there and we can merge this awesome PR! 😁

src/dev/impl/DevToys/Views/Tools/Converters/Timestamp/TimestampToolPage.xaml.cs
niyari: I totally missed it!
I set the breakpoints but could not reach them, so I left it at that.

In the current main branch code, what conditions go into the if?
It looks like it will have date text in the future, but right now I think it is only numerical.

veler (approved):
It looks good now! :D Thank you very much for this great pull request! I appreciate your contribution a lot! :D

Issue #483, opened 2022-04-04
Unix timestamp converter


KillyMXI:
### What's the Problem?

Once I opened the Unix timestamp converter - it had some value filled in.
But after I start fiddling with it - there is no way to reset the value to some useful defaults.

Most useful values I might need are going to be spread around current time, not the start of the epoch.

### Solution/Idea

"Now" button that will set the local date and time according to the clock.

It will be much easier to dial in values like "X units of time away from now".

### Alternatives

none

### DevToys Version

Version 1.0.5.0 | X64 | RELEASE | 99757dc | 99757dc

PR #515, opened 2022-05-08
#483 Added Now button to bring timestamp to current date and time.

MattWood21:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Introduces request from #483 which asks to have a Now button. The button could also be labeled "Reset", but the term "Now" is more explicit about what the button intends to do (set all times to the current date and time).

Issue Number: #483 

## What is the new behavior?

- Adds a "Now" button in the Timestamp converter
- Clicking the "Now" button sets time to currnet date and time and resets the display window times
- Behaves similar to how the page is initiated
- Added default transactions for "Now" button

## Other information

image

## Quality check

### DIFF
--- a/src/dev/impl/DevToys/Strings/fr-FR/Common.resw
+++ b/src/dev/impl/DevToys/Strings/fr-FR/Common.resw
@@ -168,6 +168,9 @@
   <data name="Megabytes" xml:space="preserve">
     <value>MO</value>
   </data>
+  <data name="Now" xml:space="preserve">
+    <value>À présent</value>
+  </data>
   <data name="OpenFile" xml:space="preserve">
     <value>Ouvrir un fichier</value>
   </data>
--- a/src/dev/impl/DevToys/Strings/fr-FR/Common.resw
+++ b/src/dev/impl/DevToys/Strings/fr-FR/Common.resw
@@ -169,7 +169,7 @@
     <value>MO</value>
   </data>
   <data name="Now" xml:space="preserve">
-    <value>À présent</value>
+    <value>Now</value>
   </data>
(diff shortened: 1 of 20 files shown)


src/dev/impl/DevToys/Strings/fr-FR/Common.resw
```diff
@@ -168,6 +168,9 @@
   <data name="Megabytes" xml:space="preserve">
     <value>MO</value>
   </data>
+  <data name="Now" xml:space="preserve">
+    <value>À présent</value>
```
veler, line 172: I'd suggest to keep the word in English for language that you may not be fluent with, and let people who are comfortable with the language translating it. Here is an example: while `À présent` is correct literal translation of `Now`, in this context, it's incorrect and is confusing. The correct one would be `Maintenant`.

veler (changes requested):
Hello,
Thanks for this contribution. It looks good overall. Can you please change the text in the .resw that have been translated through a tool like Google Translate to keep it in English (`Now`) ? We can then let translators, who are better than Google Translate, translate the language correctly.

Thank you :)

src/dev/impl/DevToys/Strings/fr-FR/Common.resw
MattWood21: Fixed in [cbf2329](https://github.com/veler/DevToys/pull/515/commits/cbf2329c2a35103e35fc16bb30ba6228c47445ad).

I wasn't sure if English or Google Translate was preferred in contributions. Now I know. 👍

veler (approved):
it looks great now! :D Thank you very much for this !

Issue #526, opened 2022-05-16
[Base64 Image] Click on save or view while no input makes the app crashes


SiderealArt:
### Description

While using Base64 Image Encoder/Decoder, the app will crash if you don't provide a base64 string and click on the save or view button.

### Steps To Reproduce

1. Switch to Base64 Image Encoder/Decoder
2. Click the save button or view button on the right.

### Expected behavior

Remind the user to provide a base64 image or disable the button before the image is pasted.

### DevToys Version

Version 1.0.7.0 | X64 | RELEASE | bbde53a | bbde53a

### Relevant Assets/Logs

![age](https://user-images.githubusercontent.com/30827929/168521483-c2f10b57-95e3-4669-b179-8cfa39c900af.gif)

PR #529, opened 2022-05-17
[Base64 Image] Click on save or view while no input made the app crashing

veler:
## Pull request type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #526 

## What is the new behavior?

The buttons are now hidden if there's no image visible.

## Quality check

- [ ] Followed the code style guideline as described in [CONTRIBUTING.md](https://github.com/veler/DevToys/blob/main/CONTRIBUTING.md)

### DIFF
--- a/src/dev/impl/DevToys/UI/Controls/ImagePreview.xaml
+++ b/src/dev/impl/DevToys/UI/Controls/ImagePreview.xaml
@@ -1,15 +1,17 @@
 ﻿<UserControl
     x:Class="DevToys.UI.Controls.ImagePreview"
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
-    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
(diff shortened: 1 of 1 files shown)


btiteux:
Hello,

I just tested the pull request, there is no button if no image that's good.

But I tried with a picture (around 578Kb) the picture is converted to base64, but then I can't do anything. 
image

I had to wait almost two minutes to get a reaction from the tools an idea to avoid that?

veler:
> Hello,
> 
> I just tested the pull request, there is no button if no image that's good.
> 
> But I tried with a picture (around 578Kb) the picture is converted to base64, but then I can't do anything. image
> 
> I had to wait almost two minutes to get a reaction from the tools an idea to avoid that?

Yup. It's a totally unrelated issue. The problem here is the UWP TextBox. It doesn't handle well very long text that doesn't have any whitespace :'(
Best alternative I can think about is using the Monaco editor, but it sounds a bit overkill too, and not ideal at the end of the day. What do you think?

veler:
Following private conversation => we will use Monaco editor and will address it in another PR.

Issue #292, opened 2022-01-28
Cron expression to human readable description and back


filipsky:
### Solution/Idea

I would love to have some cron expression to human readable format convertor

something like [this page](https://www.freeformatter.com/cron-expression-generator-quartz.html) does

### Alternatives

nothing

### Priorities

| Capability | Priority |
| :---------- | :------- |
| Cron expression show description | Must |
| Cron expression editor | Could |

krynczak:
It would be great if we could chose between Unix like syntax (5 fields: minute, hour, day of the month, month, day of the week) and other implementation syntax (ie: [aws implementation](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-create-rule-schedule.html#eb-cron-expressions) with 6 fields: minutes, hours, day-of-month, month, day-of-week, year)

DotNetInAction:
> It would be great if we could choose between Unix like syntax (5 fields: minute, hour, day of the month, month, day of the week) and others implementation syntax (ie: [aws implementation](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-create-rule-schedule.html#eb-cron-expressions) with 6 fields: minutes, hours, day-of-month, month, day-of-week, year). I'm not sure there are more implementations though

PR for this tool is now ready. I have built in a feature to switch between 5 and 6 segment Cron

PR #535, opened 2022-05-21
feat: add cron parser tool

DotNetInAction:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Currently there is no tool to parse, validate and create Cron expressions in DevToys

Issue Number: #292

## What is the new behavior?

-New tool called Cron Parser has been added under the Generators section
-New tool allows user to create and validate Cron expression to see upcoming scheduled dates
-New tool supports 5 and 6 segment Cron (Without and With seconds), has ouput line limiter and Output date time formater with validation
-New tool runs on Local timezone to show user the correct output 
-It supports daylight saving's
-It is based on open-source library called Cronos

image

## Other information

Future modifications will include Cron Maker from the UI, cleaner explanation of Cron elements, and segments descriptions

## Quality check


veler:
Hello,
Thank you very much for this pull request. The cron tool is something that is quite asked by people and I'm sure this will please a lot of customers!

I like the UI overall. A few feedback about it:
The `Next scheduled dates` doesn't scale vertically when the window is being resized. It would be nice to support it like `Regex Tester` tool (you can take example on its implementation ;-) ) 
https://github.com/veler/DevToys/blob/main/src/dev/impl/DevToys/Views/Tools/Text/RegEx/RegExToolPage.xaml

image

Also, the UI feels crowded when we turn on the Compact Overlay (Picture in Picture) mode:
image

Perhaps we could hide the options in this scenario? It's something we do in many tools like JSON Formatter. You can use VisualState for this. Here is an example: https://github.com/veler/DevToys/blob/ca0b2f838ce5e5743d904430be0ac323a80ace6f/src/dev/impl/DevToys/Views/Tools/Formatters/JsonFormatter/JsonFormatterToolPage.xaml#L21

Thank you again :) This is a great PR !

DotNetInAction:
Thank for the review and comments!

I have update my view:

1. Bottom TextBox now fills up the empty space on any size window
2. I have made settings collapsable in compact mode

Let me know what you think!

DotNetInAction:
Sorry @veler, i thought those two UI changes are the only feedback, unless im being blind :D

veler (changes requested):
No worries. As I mentioned, I left a few other comments in the code itself. :)

src/dev/impl/DevToys/Assets/Icons/Cron.svg
```diff
@@ -0,0 +1,10 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48">
```
veler, line 1: Where is this SVG coming from? Is there a license we should be aware of?

veler (approved):
Alright, I addressed all the issues. Thanks again for this contribution. It will be available in the next update. 😁

PR #365, opened 2022-02-09
Alternative search keywords

niyari:
## Pull request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Internationalization and localization
- [ ] Other (please describe):

## What is the current behavior?

Issue Number: #351

## What is the new behavior?

- Improve the user's search experience by providing searchable words.

 
image

## Quality check

Before creating this PR, have you:

- [x] Followed the code style guideline as described in [CONTRIBUTING.md](https://github.com/veler/DevToys/blob/main/CONTRIBUTING.md)
- [x] Verified that the change work in Release build configuration
- [ ] Checked all unit tests pass

PR #547, opened 2022-05-28
fix SearchKeywords value.

niyari:
## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Internationalization and localization
- [x] Other (please describe):

Issue Number: N/A

## What is the new behavior?

- Remove unnecessary strings automatically added by external services.

## Quality check

### DIFF
--- a/src/dev/impl/DevToys/Strings/ja-JP/GZipEncoderDecoder.resw
+++ b/src/dev/impl/DevToys/Strings/ja-JP/GZipEncoderDecoder.resw
@@ -163,6 +163,6 @@
     <value>GZip を利用した文字列の圧縮と展開</value>
   </data>
   <data name="SearchKeywords" xml:space="preserve">
-    <value>Checksum</value>
+    <value></value>
   </data>
 </root>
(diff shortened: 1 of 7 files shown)


niyari:
I assume this is a Crowdin specification, but when the original text is empty, some keys cannot be edited by the translator and some keys can.
(This will not be a problem for many projects because it prevents translation of sentences that are empty.)
image

Is there a setting on the Crowdin side that can be set to allow or disallow editing?
This is a suggestion to deal with the case where there is no such setting.
When implementing Search Keywords ( #365 ), the system is designed so that strings already included in tool names, etc. are not affected.
By putting the tool name, etc. in the en-US value, the translator can edit it.

veler:
Indeed, that's the issue. Good catch! I wonder however if we should fix that directly in Crowdin instead?

niyari:
Yes, the file in the current pull request is the one that should be edited using Crowdin. 
Files that could not be modified from Crowdin are included in the pull request.

Other than PR files, SearchKeywords in the LoremIpsumGenerator.resw file is empty in en-US, but I was able to change it in Crowdin, from `Checksum` to `仮文章`.
If words related to en-US SearchKeywords are prepared, they can be translated.(probably.)

veler:
Gotcha, thank you very much :)

Issue #516, opened 2022-05-09
HMAC-SHA256 Generator Tool


wapenshaw:
### What's the Problem?

Could we have under hash Generators the HMAC-SHA256/512 generator?

### Solution/Idea

[HMAC(Hash-based message authentication code)](https://en.wikipedia.org/wiki/HMAC) is a message authentication code that uses a cryptographic hash function such as SHA-256, [SHA-512](https://www.slideshare.net/sharifulr/secure-hash-algorithm-sha512) and a secret key known as a cryptographic key. HMAC is more secure than any other authentication codes as it contains Hashing as well as MAC.

### Alternatives

Its a simple tool previewed here https://www.devglan.com/online-tools/hmac-sha256-online

### Priorities

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow developers to accomplish generating Hash-based message authentication code using a secret key |

### Comments

Ability to generate hash based on a secret key

L1nu5:
I thought of this design: A simple text box for the secret key - if it's non-empty, then use HMAC variants (`HMACMD5, HMACSHA1` etc), otherwise simply use regular hashing algorithms. I have a prototype in progress, I'll share screenshots in few days.

L1nu5:
Here's the functional version that I could implement: 

![image](https://user-images.githubusercontent.com/14090904/168972234-4d74639e-2e25-4c39-bf04-7f478cd92240.png)

There are other designs which are also possible:
1. Separate Hashing page for HMAC entirely (this could be an overkill)
2. A toggle switch to enable/disable Secret key textbox

I'm using `MacAlgorithmProvider` currently for this feature. I also noticed that SHA 384 hash output is missing, I could add implement that as well but I'm not sure why it was not added on the Hash generator screen.

wapenshaw:
@L1nu5 I agree with both your points, separate page just for HMAC would be overkill, by default you can disable the HMAC toggle so that it wont confuse users already using this page extensively. Also the Input field for the secret key could be better off being a single line text box. 

Also thanks for responding. :)

L1nu5:
@wapenshaw No problem! Here's what it looks like now:

**1. Normal mode**

![image](https://user-images.githubusercontent.com/14090904/169394904-11593a57-5b9f-4581-a993-936b7762e669.png)

**2. HMAC mode**

![image](https://user-images.githubusercontent.com/14090904/169394968-3068c075-5ff4-4ef0-8c86-9f3485ccb8ed.png)

[...]

PR #555, opened 2022-06-03
Added HMAC authentication mode in Hash generators

L1nu5:
## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Currently, we do not have HMAC generation feature in the tool. Please refer: [516](https://github.com/veler/DevToys/issues/516)

Issue Number: [516](https://github.com/veler/DevToys/issues/516)

## What is the new behavior?

I'm using `MacAlgorithmProvider` currently for this feature. I also noticed that SHA 384 hash output is missing, I could implement that as well but I'm not sure why it was not added on the Hash generator screen.

Here's how it'll look like with the fix:
1. Normal mode

Normal_mode

2. HMAC mode

Hmac_mode

I'm currently using hardcoded strings for the toggle switch and the secret key textbox title. I just wanted to make sure that the current strings are good enough in this PR review, before I initiate a translation on Crowdin.

## Quality check


veler:
Hi,
Thank you for offering this change. I like the idea overall :D
One suggestion / question though: how do you feel about not having a `HMAC Mode` option, and simply applying the HMAC encryption when `Secret Key` isn't empty?

L1nu5:
@veler  I had initially implemented it that way but thought that the `Secret Key` text field might need some concise hint/tooltip on leaving it empty if the user only needs Hash instead of HMAC. With toggle switch, it feels slightly easier to operate without needing to explain with a tooltip/hint. I can definitely change it back without needing to use a toggle switch though and it will drastically reduce the overall LoC for the change, please let me know your thoughts :)

veler:
> @veler I had initially implemented it that way but thought that the `Secret Key` text field might need some concise hint/tooltip on leaving it empty if the user only needs Hash instead of HMAC. With toggle switch, it feels slightly easier to operate without needing to explain with a tooltip/hint. I can definitely change it back without needing to use a toggle switch though and it will drastically reduce the overall LoC for the change, please let me know your thoughts on this :)

That is fair :-) Let's keep it the way it is then !

src/dev/impl/DevToys/Views/Tools/Generators/HashGenerator/HashGeneratorToolPage.xaml
```diff
@@ -42,6 +42,17 @@
                     <ComboBoxItem Tag="Base64" Content="{x:Bind ViewModel.Strings.OutputBase64}"/>
                 </ComboBox>
             </controls:ExpandableSettingControl>
+            <controls:ExpandableSettingControl
+                Title="HMAC Mode">
```
veler, line 46: Please globalize this string by adding a new entry in https://github.com/veler/DevToys/blob/main/src/dev/impl/DevToys/Strings/en-US/HashGenerator.resw . Then open https://github.com/veler/DevToys/blob/main/src/dev/impl/DevToys/LanguageManager.tt in Visual Studio, press Ctrl+S (even if there's no changes). This will re-regenerate https://github.com/veler/DevToys/blob/main/src/dev/impl/DevToys/LanguageManager.cs automatically and you will then be able to use the resource in the XAML.

src/dev/impl/DevToys/Views/Tools/Generators/HashGenerator/HashGeneratorToolPage.xaml
```diff
@@ -50,6 +61,13 @@
             AcceptsReturn="True"
             Text="{x:Bind ViewModel.Input, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}"/>
 
+        <controls:CustomTextBox x:Name="SecretKeyInput"
+            Header="Secret Key (HMAC only)"
```
veler, line 65: Same thing here. Also, perhaps we can remove `(HMAC only)` since this field only appear when `HMAC mode` is enabled.

src/dev/impl/DevToys/ViewModels/Tools/Generators/HashGenerator/HashGeneratorToolViewModel.cs
```diff
@@ -164,12 +199,23 @@ private async Task<string> CalculateHashAsync(string alrogithmName, string text)
 
             try
             {
-                var algorithmProvider = HashAlgorithmProvider.OpenAlgorithm(alrogithmName);
-
-                IBuffer buffer = CryptographicBuffer.ConvertStringToBinary(text, BinaryStringEncoding.Utf8);
-                buffer = algorithmProvider.HashData(buffer);
-
                 string? hash = "";
+                IBuffer? buffer = null;
```
veler, line 203: Perhaps we could add some unit tests to https://github.com/veler/DevToys/blob/main/src/tests/DevToys.Tests/Providers/Tools/HashGeneratorTests.cs

veler (approved):
It looks all good ! :D Thank you so much for this contribution!!

L1nu5:
> It looks all good ! :D Thank you so much for this contribution!!

No problem, I'm really happy to contribute :)

PR #771, opened 2023-03-23
Added Smart Detection & partially Clipboard API

veler:
## Pull request type

- [ ] Bugfix
- [ ] Build related changes

DevToys 1.0 has a design flaw in its Smart Detection feature: if multiple tools support JSON data (for example), then we will try to detect JSON as many times as there are tools supporting this kind of data. This is a problem if the data in the clipboard is large, as we will have to do this work of detection twice on a very large data, which can therefore be slow.

We also got feedback that large data from the clipboard may freeze DevToys because Smart Detection runs switch back and forth on the UI Thread and end up doing a good part of the detection work on the UI thread instead of in background.

## What is the new behavior?

In DevToys 2.0:
- I made sure that the detection work happens in a background thread and won't freeze the UI even if it takes many seconds to run.
- I changed the design so data detection is run only once. Here is how it works:
    1. Tools aren't responsible for detecting the data themselves anymore. Now, data detection is done independently from the tools. It is done by implementing a `IDataTypeDetector`
    2.  Each implementation of `IDataTypeDetector` should be dedicated to 1 type of data to detect. For example, `text`, `json`, `jwt-header`.
    3. Data can depend on another. For example, we could have `IDataTypeDetector` for the following data:
        -  `text`: the `IDataTypeDetector` would simply detect whether the data is a `string` and pass this string in result.
        - `json`: would get the `string` from the previous data detector and create (for example) a `JToken` that represents the JSON (admitting the data is a valid JSON).
        - `jwt-header`: would get the `JToken` from the previous data detector and check whether it's a valid JWT Header or not.
    4. Following the example above, admitting that the data is NOT a JSON, the `jwt-header` detector won't be invoked.
- With the detectors implemented, DevToys' Tools can then use the `AcceptedDataTypeName` attribute to tell DevToys that they can accept `json`, `text`, `jwt-header` or whatever as an input. When this happen, `IGuiTool.OnDataReceived` method would be invoked with the detected data in argument. The tool will then pass this data to its UI, or do whatever it wants with it.

When using SmartDetection service, there are 2 modes of detection available:
1. `strict`
In this mode, admitting that the data is a JSON, only tools supporting a JSON will be detected. Tools supporting plain text won't be returned by the service. We'd use this mode when detecting the content of the clipboard when opening DevToys to suggest tools in the NavigationView that match the best the clipboard.
2. `non-strict`
In this other mode, admitting that the data is still a JSON, tools supporting JSON and plain text will be detected. Tools supporting JSON would be at the top of the list of tools returned by the service. We'd use this mode inside of the CodeEditor control to offer the user to transfer a data from a tool to another, depending on the detected data. For example, when decoding a Base64 text, if the decoded text is a JSON, the CodeEditor could then have a UX to suggest to transfer this JSON to the JSON Formatter tool (top of the list), or the Text Diff tool (bottom of the list).
 
## Other information

I also started to add a `IClipboard` service for WASDK and MAUI app. The MacOS implementation is missing for the moment.
We will be able to support Text, File List and Images, which differs from DevToys 1.0 which could only smart detect Text.

image

## Quality check

Before creating this PR:

   - [X] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)


src/app/dev/DevToys.Tools/Tools/Sample/SampleToolFactory.cs
```diff
@@ -42,12 +41,13 @@ public ValueTask<int> InvokeAsync(ILogger logger, CancellationToken cancellation
...
     GroupName = "Encoders / Decoders",
     ResourceManagerAssemblyIdentifier = nameof(DevToysToolsResourceManagerAssemblyIdentifier),
-    ResourceManagerBaseName = "DevToys.Tools.Sample.Sample",
+    ResourceManagerBaseName = "DevToys.Tools.Tools.Sample.Sample",
     ShortDisplayTitleResourceName = nameof(Sample.CommandDescription),
     LongDisplayTitleResourceName = nameof(Sample.LongTitle),
     DescriptionResourceName = nameof(Sample.Description),
     AccessibleNameResourceName = nameof(Sample.CommandDescription),
     SearchKeywordsResourceName = nameof(Sample.CommandDescription))]
+[AcceptedDataTypeName(PredefinedCommonDataTypeNames.Json)]
```
veler, line 50: @btiteux , here is how a tool can tell that it accepts a JSON as an input from Smart Detection

src/app/dev/DevToys.Business/ViewModels/MainWindowViewModel.cs
```diff
@@ -154,6 +172,71 @@ internal void UpdateWindowTitle(bool isInCompactOverlayMode)
         }
     }
 
+    /// <summary>
+    /// Get the data from the clipboard and try to detect tools that can accept the clipboard data as an input.
+    /// </summary>
+    internal async Task RunSmartDetectionAsync(bool isInCompactOverlayMode)
+    {
```
veler, line 179: Something I should improve here:

Smart Detection happens in a background thread and doesn't freeze the UI anymore. That's great! However it may still take 5-10 seconds, and there's no indication of what's going on, and Smart Detection may navigate the user automatically to a tool we detected when done.

This can be troublesome for the user experience. Here are things we could do to improve the UX (but let's do that in a separated PR?) :

1. Set a timeout so we don't spend more than 2-3 seconds doing the detection.
2. if user navigated somewhere by himself since we started the detection, cancel the detection, because it sounds like the user knowns where he's going.

Thoughts @btiteux  ?

src/app/dev/DevToys.Core/Tools/SmartDetectionService.cs
```diff
@@ -0,0 +1,250 @@
...
+
+        /// <summary>
+        /// Gets the detector associated with this node.
+        /// </summary>
+        internal Lazy<IDataTypeDetector, DataTypeDetectorMetadata> Detector { get; }
+
+        /// <summary>
+        /// Gets the list of child detectors for this node.
+        /// </summary>
+        internal List<DetectorNode> ChildrenDetectors { get; } = new();
```
veler, line 248: Some node may not have any child. Let's make this property nullable and only allocate the list if we're actually adding an item?

src/app/dev/platforms/desktop/DevToys.MauiBlazor/Core/Clipboard/Clipboard.cs
```diff
@@ -0,0 +1,62 @@
...
+using Windows.Storage.Streams;
+
+namespace DevToys.MauiBlazor.Core.Clipboard;
+
+[Export(typeof(Api.Core.IClipboard))]
+internal sealed partial class Clipboard : Api.Core.IClipboard
+{
+    public async Task<object?> GetClipboardDataAsync()
+    {
+        if (Microsoft.Maui.ApplicationModel.DataTransfer.Clipboard.Default.HasText)
```
veler, line 12: I should add a try /catch here like for the WASDK app.

veler: Done

src/app/dev/DevToys.Core/Tools/SmartDetectionService.cs
veler: Done

src/app/dev/DevToys.Business/ViewModels/MainWindowViewModel.cs
veler: Done

btiteux (approved):

Issue #494, opened 2022-04-15
Text case converter and inspector only shows 12108 lines when pasting a really long string


DavidCZ2051:
### Description

The "Text case converter and inspector" only shows 12108 lines but still counts the information right when pasting a long string.

### Steps To Reproduce

1. Go to the "Text case converter and inspector".
2. Paste in a long string that has over 12108 lines.
3. Wait for it to process.

### Expected behavior

Should show all the lines pasted by the user.

### DevToys Version

Version 1.0.6.0 | X64 | RELEASE | c032ebb | c032ebb

### Relevant Assets/Logs

![screenshot](https://user-images.githubusercontent.com/84991177/163591638-636aba29-7f7a-4a06-9c45-774be246c907.png)

L1nu5:
I'll start working on this one.

L1nu5:
I think there is another issue here- DevToys is frozen for the entire duration of the copy paste of such huge text data. I noticed that memory usage is well beyond 1GB. For me, I could not use DevToys after pasting the text and here are the statistics of the loaded text

![image](https://user-images.githubusercontent.com/14090904/163658205-4af9057e-e778-4b92-9c74-be6ae1729959.png)

![image](https://user-images.githubusercontent.com/14090904/163657993-259f1432-c33a-4f59-aac6-896122bfb51d.png)

I could think of two ways to fix this, but this could be a new feature itself:
1. Have a file open dialog like feature instead of having to copy/paste this much text data (optional)
2. Show a dialog box with a progress bar and a cancel button when data is being copied to the text input

niyari:
This phenomenon is related to the `CustomTextBox` drawing process.
Issue #275 was (temporarily) solved by setting the output to a maximum of 10,000 lines.

veler:
Yeah, it's unfortunately a known limitation of UWP's TextBox :( It doesn't scale well with very large texts... Code Editor (precisely the monaco-editor used in DevToys) has a better tolerance, but as you likely noticed, it takes about 1 second to load on each tool. If we use it on absolutely every tool, I'd find it a bit sad.

I guess another possibility is to re-implement ourselves a fully functional text box, but that's a very long and challenging work actually.

As of today, I don't have an ideal solution to bring. The best is likely the Code Editor, but again, not ideal...

veler:
Fixed in DevToys 2.0

PR #780, opened 2023-04-01
Added a multiline-line text input

veler:
## Pull request type

- [ ] Bugfix
- [ ] Build related changes

DevToys 2.0 doesn't have a multi-line text input yet.

## What is the new behavior?

Added a multi-line text input control that supports read-only and read-write mode.
Compared to DevToys 1.0, this time, every multi-line text input will use the Monaco Editor. This will help DevToys handling scenario where users paste very long text where we weren't necessarily expecting it.

DevToys 1.0 has a memory consumption issue related to Monaco: every time the monaco editor is needed, we create and allocate a new one. That means that when jumping from a tool that use Monaco to another that also use Monaco, we will now have 2 instances of Monaco (with 2 WebView) in memory. Each instances of Monaco consumes roughly 150 MB of RAM. This makes DevToys quickly consuming GB(s) of RAM.
In addition, loading the Monaco Editor is slow. It requires us to display a Progress Ring every time we open a tool for the first time that needs the editor.

In DevToys 2.0, we now use a pool of Monaco instances. Every time we jump from a tool to another, we will "release" the Monaco Editor that was in use in the tool, detach it from the UI. When navigating to the other tool, we will ask the pool to provide us an instance of Monaco. The pool will either provide an instance that got previously released, or produce a new one.

This way, when 2 tools need 1 Monaco editor, we allocate only one.
When navigating to the new tool, we change the text, language, settings of that Monaco instance we reused, so it adapts to the new tool. There might be a short time where we see the Monaco editor with a stale state, but in my opinion this constraint worth the gain of memory usage and loading speed of the tool.

This PR also address the following issues: #755, #494

## Other information

https://user-images.githubusercontent.com/3747805/229258383-c70ef0a1-3267-4361-8767-2940729c4dc9.mp4

## Quality check

Before creating this PR:

   - [X] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)


veler:
Last commit added support of fonts. Here, Cascadia Mono is used after being detected as the best font to use.

image

image

Issue #758, opened 2023-02-20
URL Encoder should treat newline to %0D%0A


doggy8088:
### Description

![image](https://user-images.githubusercontent.com/88981/220060949-865bbaa4-c488-483d-89f0-4ca00f13b53f.png)

### Steps To Reproduce

1. Go to 'Encoders / Decoders > URL'
2. Choose 'Encode'
3. Input: 

    ```
    123
    123
    ```

4. Output will be `123%0D123` which is wrong. It should be `123%0D%0A123` or `123%0A123`

### Expected behavior

It should be `123%0D%0A123` or `123%0A123`.

### DevToys Version

Version 1.0.10.0 | X64 | RELEASE | b949a18 | b949a18

### Relevant Assets/Logs

```txt
# - [02/20/2023 17:07:53]
Feature name: Failed to copy from custom text box
Custom message: 
Exception message: OpenClipboard Failed (Exception from HRESULT: 0x800401D0)
Exception stack trace:
   at System.Runtime.InteropServices.McgMarshal.ThrowOnExternalCallFailed(Int32, RuntimeTypeHandle) + 0x21
   at __Interop.ComCallHelpers.Call(__ComObject, RuntimeTypeHandle, Int32) + 0xb8
   at __Interop.ForwardComStubs.Stub_17[TThis](__ComObject, Int32) + 0x24
   at Windows.ApplicationModel.DataTransfer.Clipboard.Flush() + 0x2c
   at DevToys.UI.Controls.CustomTextBox.CopyTextBoxSelectionToClipboard() + 0xc6
```

pulsarleaf:
From my testing, it seems the inherent C# functions used for the URL Encoder have no option for it to return anything other than a %0D. It would require a custom function to either do the whole encoding or to scan for %0D and add %0A after or replace it.

ebejko:
Could be related to this issue
https://github.com/microsoft/microsoft-ui-xaml/issues/1826

doggy8088:
I think @ebejko is right.  It's related to textbox control. Any text put it on that control. All the newline can only be interpret as `\r` only. It's so weird.

No matter what this issue should be addressed and it's fundamental. This should have some way to solve it.

PR #789, opened 2023-04-05
Adding url encoding wrapper function to handle xaml textbox conversio…

jlevier:
…n of new line to only carriage return

## Pull request type

- [x] Bugfix
- [ ] Build related changes

Issue Number: #758 

## What is the new behavior?

- Due to https://github.com/microsoft/microsoft-ui-xaml/issues/1826, it seems that the xaml textbox is converting a new line to a single carriage return.  This resulted in the url encoding not properly handling the user input.  This change will support if the decision is made to convert to carriage return + line feed or just line feeds.

## Other information

screenshot

## Quality check

Before creating this PR:

- [x] Did you check for accessibility? On Windows, you can use Accessibility Insights for this.
   - [x] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)

### DIFF
--- a/src/dev/impl/DevToys/Helpers/UrlHelper.cs
+++ b/src/dev/impl/DevToys/Helpers/UrlHelper.cs
@@ -1,0 +1,22 @@
+﻿using System;
+
+namespace DevToys.Helpers
+{
+    internal static class UrlHelper
+    {
+        /// <summary>
+        /// Url encodes a string.
+        /// This is a wrapper function for the built in System.Uri api to accommodate this issue:
+        /// https://github.com/microsoft/microsoft-ui-xaml/issues/1826
+        /// </summary>
+        /// <param name="data">Input to url encode</param>
+        /// <returns>Url encoded result</returns>
+        internal static string UrlEncode(string data)
+        {
+            string newLineAdjusted = data.Replace("\r\n", "\n").Replace("\r", "\n");
+            string encoded = Uri.EscapeDataString(newLineAdjusted);
+
+            return encoded;
+        }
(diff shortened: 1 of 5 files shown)


src/dev/impl/DevToys/Helpers/UrlHelper.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
```
veler, line 2: Are these necessary? I think only `System` is needed here.

veler (changes requested):

src/dev/impl/DevToys/Helpers/UrlHelper.cs
jlevier: @veler , thank you good catch.  Would you be interested in adding 
```bash
dotnet_diagnostic.IDE0005.severity = error
```
to the `.editorconfig`? That way it would be clear to see:
image

jlevier:
Ok, sounds great.  Thank you so much for the speedy review!

src/dev/impl/DevToys/Helpers/UrlHelper.cs
veler: Thanks for the suggestion. I'd rather avoid. In DevToys 2.0 branch, at least for now, we have some `using` that are wrongly marked as unnecessary, so I'm concerned it would be cause more pain for now. But I will keep that in mind for future evolution. :)

veler (approved):

PR #791, opened 2023-04-07
MAUI Blazor sidebar and fluent ui base design

btiteux:
## Pull request type

- [ ] Bugfix
- [ ] Build related changes

Devtoy 2.0 doesn't have blazor component and fluent design

Issue Number: N/A

## What is the new behavior?

Added a blazor Navbar component for .net MAUI.

## Other information

image
image

## Remaining TODO: 
- Use Event from ViewModel.
- Some CSS adjustment
- Search form

## Quality check

Before creating this PR:

   - [x] Windows
   - [x] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)


.vscode/launch.json
```diff
@@ -1,38 +1,57 @@
...
+            "processName": "DevToys.MauiBlazor",
+            "preLaunchTask": "Run",
+        }
+    ]
+}
+
+
+
+
+// {
```
veler, line 20: Oops, let's remove this? XD

.vscode/tasks.json
```diff
@@ -1,99 +1,186 @@
...
+            },
+            // Use the standard MS compiler pattern to detect errors, warnings and infos
+            "problemMatcher": ["$msCompile"]
+        }
+    ]
+}
+
+
+
+// ======
```
veler, line 87: This too ^^

src/app/dev/platforms/desktop/DevToys.MauiBlazor/Components/Navbar/Navbar.razor.cs
```diff
@@ -0,0 +1,99 @@
...
+    [Parameter]
+    public bool Collapsed { get; set; } = false;
+
+    public string CollapsedName => Collapsed.ToString().ToLowerInvariant();
+
+    [Parameter]
+    public RenderFragment? ChildContent { get; set; }
+
+    [Parameter]
+    public EventCallback<NavbarItem> CurrentSelectedChanged { get; set; }
```
veler, line 39: Looks like this isn't used. Should we remove it or will it be used soon?

src/app/dev/platforms/desktop/DevToys.MauiBlazor/Components/Navbar/Navbar.razor.cs
```diff
@@ -0,0 +1,99 @@
...
+
+    [Parameter]
+    public EventCallback<NavbarItem> OnSelectedChange { get; set; }
```
veler, line 42: Same here

src/app/dev/platforms/desktop/DevToys.MauiBlazor/Components/Navbar/NavbarAnchor.razor.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    protected override void OnParametersSet()
+    {
+        string? href = Href;
+        if (AdditionalAttributes != null && AdditionalAttributes.TryGetValue("href", out object? obj))
+        {
+            href = Convert.ToString(obj, CultureInfo.InvariantCulture);
+        }
+
+        _hrefAbsolute = href == null ? null : NavigationManager.ToAbsoluteUri(href).AbsoluteUri;
```
veler, line 66: You set `_hrefAbsolute ` to null but `_hrefAbsolute ` isn't marked as nullable. Should this be `string.Empty` instead?

src/app/dev/platforms/desktop/DevToys.MauiBlazor/Shared/MainLayout.razor
```diff
@@ -1,23 +1,58 @@
...
+            else
+            {
+                <li class="nav-item nav-separator"></li>
+            }
+        }
+    </NavbarItems>
+</Navbar>
+
+<main class="page">
+    @Body
```
veler, line 57: I'm playing a bit with the code to try to make the components for the tools. I notice that for now we only have `Index` and it uses the URI to determine the tool name to display.

Thinking about it, I thought `OK but we don't have only tools, sometimes we also display a list of tools in this area, which isn't a tool itself`. So naive as I am, I started making a page `ToolPage.razor` and `ToolGroupPage.razor`, exactly like in the WASDK app. One page would have the URI `tool/{Name}` and the other `group/{Name}` but then I quickly realized that you did it this way because when clicking on a menu item, the whole page would refresh when navigating to a different URL, which is a problem indeed....

I'm not fan of the idea of mixing the Tool UI and the Group UI in the same page (`Index.razor`).

So how about the following:
Instead of navigating to a page, we listen to a `MainWindowViewModel.SelectedMenuItem` and it change, depending on whether it's a `GuiToolViewItem` or `GroupViewItem`, we show here a `ToolPage` component or a `ToolGroupPage` component. This way we never navigate anywhere and just replace the body on the fly. In addition, this way we pass the whole `GuiToolViewItem` to the `ToolPage` instead of relying on the internal name to find which tool / group we're trying to display.

Would this work? Would this have any performance impact compared to navigating (which, if I understood well, isn't really navigating at the moment since we don't refresh the whole page, right?) ?
What do you think? :)

veler: PS: This could be done in a different PR. Happy to do it myself too if you want, since I kind of need this first before creating the tool's UI component.

src/app/dev/platforms/desktop/DevToys.MauiBlazor/Components/Navbar/NavbarAnchor.razor.cs
```diff
@@ -0,0 +1,105 @@
...
+        _isActive = ShouldMatch(NavigationManager.Uri);
+        if (_isActive)
+        {
+            ClassesHasChanged();
+            StateHasChanged();
+        }
+        base.OnParametersSet();
+    }
+
+    protected async Task OnClickHandler(MouseEventArgs eventArgs)
```
veler, line 76: `OnClickAsync` ?

src/app/dev/platforms/desktop/DevToys.MauiBlazor/Components/Navbar/NavbarGroup.razor.cs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    protected override void AppendClasses(ClassHelper helper)
+    {
+        helper.Append("nav-group");
+        base.AppendClasses(helper);
+    }
+
+    // Todo Handle collapse animation
+    internal async Task OnClickHandler()
```
veler, line 62: `OnClickAsync` too ?

PR #834, opened 2023-06-04
[DevToys 2.0] Added context menu and popup system.

veler:
## Pull request type

- [ ] Bugfix
- [ ] Build related changes

We need to add context menu, drop down list and teaching-tips flyout in DevToys Blazor implementation.

## What is the new behavior?

This PR contains the following new components:
1. **Render**: Used for rendering anything given inside and programmatically trigger a refresh by exposing `StateHasChanged` as a public method. This is used by the `Popover` below.

2. **Popover** and **PopoverProvider**: Forked from [MudBlazor](https://mudblazor.com/components/popover#simple-popover), this is an approach that enables displaying popup anywhere in the UI (see demo on MudBlazor website). It can be sticked to another component, like a `TextBox` or `Button`. This is currently used to enable `ContextMenu` (see below). But it will also be used to enable the `AutoSuggestBox` (the search bar displayed in the NavBar that shows a drop down list as we type).

4. **Overlay**: This is also forked from MudBlazor. It can be used for various purposes. In this PR, we use it to prevent clicking anywhere on the window while a context menu is opened. If a click happens on the Overlay, the context menu gets closed.

5. **FocusTrapper**: homemade component that traps the keyboard focus within this component, ensuring that pressing Tab or Shift+Tab won't move the focus outside of this component.

6. **ContextMenu**: Finally, the actual context menu that displays items in a `FocusTrapper`, in a `Popover`, and trigger a global `Overlay` to be displayed.

image

## Remains to do:

I set a bunch of `// TODO` in the code. A few things that remain to do is, for example, placing the context menu items inside of a list that supports Up/Down arrow keys to select the item of our choice. I believe we could make a `ListBox` component that allows this mechanism (we need one to make some tools) and reuse it in the context menu.

Another thing to do is to detect when the window is moved, resized or lose focus and trigger the context menu to close when this happen.

## Quality check

Before creating this PR:

- [X] Did you check for accessibility? On Windows, you can use Accessibility Insights for this.
   - [X] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)


src/app/dev/DevToys.Blazor/Components/Popover/Origin.cs
btiteux, line 1: Hello, if we want to retrieve the description, it could be better to use an Enumeration Class like we did in the v1. This way we can avoid the reflection on the attribute ?

src/app/dev/DevToys.Blazor/Components/Popover/OverflowBehavior.cs
btiteux, line 1: Hello, same point here as in Origin.cs

src/app/dev/DevToys.Blazor/Core/Extensions/EnumExtensions.cs
btiteux, line 1: Hello, no need for this extension if we use the Enumeration Class :)

Issue #232, opened 2022-01-24
Sort lines of text alphabetically


phytemit:
### What's the Problem?

Sort lines of text alphabetically

### Solution/Idea

Sort lines of text alphabetically

### Alternatives

Sort lines of text alphabetically

### Priorities

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow developers to accomplish W | Must |
| This proposal will allow end users to accomplish X | Should |
| This proposal will allow developers to accomplish Y | Could |
| This proposal will allow end users to accomplish Z | Won't |

RicardoNiepel:
I would love to see this - maybe we should make the Feature Request more general?

Just comes to my mind:
* Alphabetize + reverse order
* Alphabetize list by last word/name + reverse order
* Remove duplicates 
* Reverse list
* randomize

celluj34:
I use some of these tools semi-regularly: https://alphabetizer.flap.tv/

micahmo:
How would we feel about sticking this alongside the existing `StringUtiltiies`? 

https://github.com/veler/DevToys/assets/7417301/9e31476f-dd56-4e01-8edf-d6b150d4d3b1

veler:
That's what I had in mind. I love this video by the way! :D Are you cooking anything you want to push to this repository?

micahmo:
> That's what I had in mind. I love this video by the way! :D Are you cooking anything you want to push to this repository?

I'm glad you asked! 😉 #835

PR #835, opened 2023-06-05
Add line sorting to string utils

micahmo:
## Pull request type

- [ ] Bugfix
- [ ] Build related changes

Issue Number: #232

## What is the new behavior?

Add the following functions to the string utilities tool. These operate on lines of text.

- Alphabetize
- Reverse alphabetize
- Alphabetize by last word
- Reverse alphabetize by last word
- Reverse
- Randomize

https://github.com/veler/DevToys/assets/7417301/9e31476f-dd56-4e01-8edf-d6b150d4d3b1
 
## Other information

All of the line splitting is done on CR (`\r`). I noticed that this is the line separator used for typed text and loaded files using any line endings (CR, LF,"or or CRLF). I don't really understand why, but it seems consistent.

For sorting by last word, I'm splitting on space and also comma (to accommodate CSV and last,first formatting).

It might be ever so slightly confusing that the "Original text" button is under the "Convert" heading. Maybe we could consider making it a highlighted/action button.

## Quality check

Before creating this PR:

   - [x] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)

### DIFF
--- a/src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
+++ b/src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
@@ -702,13 +702,7 @@ namespace DevToys.ViewModels.Tools.StringUtilities
                 string? line1LastWord = line1.Split(' ', ',').LastOrDefault();
                 string? line2LastWord = line2.Split(' ', ',').LastOrDefault();
 
-                if (string.Equals(line1LastWord, line2LastWord, StringComparison.CurrentCulture))
-                {
-                    return 0;
-                }
-
-                var newList = new List<string> { line1LastWord, line2LastWord }.OrderBy(word => word).ToList();
-                return newList.IndexOf(line1LastWord) == 0 ? -1 : 1;
+                return string.Compare(line1LastWord, line2LastWord, StringComparison.CurrentCulture);
             });
             text = string.Join('\r', lines);
 
@@ -740,13 +734,7 @@ namespace DevToys.ViewModels.Tools.StringUtilities
                 string? line1LastWord = line1.Split(' ', ',').LastOrDefault();
                 string? line2LastWord = line2.Split(' ', ',').LastOrDefault();
 
-                if (string.Equals(line1LastWord, line2LastWord, StringComparison.CurrentCulture))
-                {
-                    return 0;
-                }
-
-                var newList = new List<string> { line1LastWord, line2LastWord }.OrderByDescending(word => word, StringComparer.CurrentCulture).ToList();
-                return newList.IndexOf(line1LastWord) == 0 ? -1 : 1;
+                return string.Compare(line1LastWord, line2LastWord, StringComparison.CurrentCulture) * -1;
             });
             text = string.Join('\r', lines);
 
--- a/src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
+++ b/src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
@@ -646,8 +646,7 @@ namespace DevToys.ViewModels.Tools.StringUtilities
                 return;
             }
 
-            var lines = text.Split('\r').ToList();
-            lines.Sort(StringComparer.CurrentCulture);
+            IOrderedEnumerable<string> lines = text.Split('\r').OrderBy(line => line, StringComparer.CurrentCulture);
             text = string.Join('\r', lines);
 
             lock (_lockObject)
@@ -672,9 +671,7 @@ namespace DevToys.ViewModels.Tools.StringUtilities
                 return;
             }
 
-            var lines = text.Split('\r').ToList();
-            lines.Sort(StringComparer.CurrentCulture);
-            lines.Reverse();
+            IOrderedEnumerable<string> lines = text.Split('\r').OrderByDescending(line => line, StringComparer.CurrentCulture);
             text = string.Join('\r', lines);
 
             lock (_lockObject)
@@ -710,8 +707,7 @@ namespace DevToys.ViewModels.Tools.StringUtilities
                     return 0;
                 }
 
-                var newList = new List<string> { line1LastWord, line2LastWord };
-                newList.Sort(StringComparer.CurrentCulture);
+                var newList = new List<string> { line1LastWord, line2LastWord }.OrderBy(word => word).ToList();
                 return newList.IndexOf(line1LastWord) == 0 ? -1 : 1;
             });
             text = string.Join('\r', lines);
@@ -749,11 +745,9 @@ namespace DevToys.ViewModels.Tools.StringUtilities
                     return 0;
                 }
 
-                var newList = new List<string> { line1LastWord, line2LastWord };
-                newList.Sort(StringComparer.CurrentCulture);
+                var newList = new List<string> { line1LastWord, line2LastWord }.OrderByDescending(word => word, StringComparer.CurrentCulture).ToList();
                 return newList.IndexOf(line1LastWord) == 0 ? -1 : 1;
             });
-            lines.Reverse();
             text = string.Join('\r', lines);
 
             lock (_lockObject)
@@ -778,8 +772,7 @@ namespace DevToys.ViewModels.Tools.StringUtilities
                 return;
             }
 
-            var lines = text.Split('\r').ToList();
-            lines.Reverse();
+            IEnumerable<string> lines = text.Split('\r').Reverse();
             text = string.Join('\r', lines);
 
             lock (_lockObject)
@@ -804,7 +797,7 @@ namespace DevToys.ViewModels.Tools.StringUtilities
                 return;
             }
 
-            var lines = text.Split('\r').OrderBy(_ => _random.Next());
+            IOrderedEnumerable<string> lines = text.Split('\r').OrderBy(_ => _random.Next());
             text = string.Join('\r', lines);
 
             lock (_lockObject)
--- a/src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
+++ b/src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
@@ -181,12 +181,7 @@ namespace DevToys.ViewModels.Tools.StringUtilities
 
         private void ExecuteOriginalCaseCommand()
         {
-            lock (_lockObject)
-            {
-                _forbidBackup = true;
-                Text = _textBackup;
-                _forbidBackup = false;
-            }
+            SetTextWithoutBackup(_textBackup);
         }
 
         #endregion
@@ -226,12 +221,7 @@ namespace DevToys.ViewModels.Tools.StringUtilities
                 }
             }
 
-            lock (_lockObject)
-            {
-                _forbidBackup = true;
-                Text = new string(sentenceCaseString);
-                _forbidBackup = false;
-            }
+            SetTextWithoutBackup(new string(sentenceCaseString));
         }
 
         #endregion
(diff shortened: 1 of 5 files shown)


src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
```diff
@@ -626,6 +634,189 @@ private void ExecuteInverseCaseCommand()
...
+        private void ExecuteReverseAlphabetizeCommand()
+        {
+            string? text = _textBackup;
+            if (text is null)
+            {
+                return;
+            }
+
+            var lines = text.Split('\r').ToList();
+            lines.Sort(StringComparer.CurrentCulture);
```
celluj34, line 676: Why `ToList` and `Sort` instead of `OrderBy`?

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
```diff
@@ -626,6 +634,189 @@ private void ExecuteInverseCaseCommand()
...
+        private void ExecuteAlphabetizeLastCommand()
+        {
+            string? text = _textBackup;
+            if (text is null)
+            {
+                return;
+            }
+
+            var lines = text.Split('\r').ToList();
+            lines.Sort((line1, line2) =>
```
celluj34, line 703: Similar to above, why not `OrderByDescending`?

src/tests/DevToys.Tests/Providers/Tools/StringUtilitiesTests.cs
```diff
@@ -529,5 +531,126 @@ string expectedResult
...
+        [DataRow("Steve Adams\rAdam,Smith\rBob Barker\rXander")]
+        public void Randomize(string input)
+        {
+            StringUtilitiesToolViewModel viewModel = ExportProvider.Import<StringUtilitiesToolViewModel>();
+
+            viewModel.Text = input;
+
+            viewModel.RandomizeCommand.Execute(null);
+
+            // Can't guarantee the order, just make sure everything is still there
```
celluj34, line 646: An interface _could_ be introduced, then we could mock and guarantee results, something like https://stackoverflow.com/a/311829/17052649 if that's acceptable.

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
micahmo: Good point, improved in 9c4d7228d8a6e92848b7f1277c343749ae964b33.

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
micahmo: Improved these and others in 9c4d7228d8a6e92848b7f1277c343749ae964b33.

src/tests/DevToys.Tests/Providers/Tools/StringUtilitiesTests.cs
micahmo: Not a bad idea. I hadn't dived into how DI is set up, so I wasn't sure if I could inject the real `Random` though the constructor. Also thought it might be a bit overkill. I guess I'll leave this one up to the maintainer.

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
```diff
@@ -626,6 +634,182 @@ private void ExecuteInverseCaseCommand()
...
+            {
+                string? line1LastWord = line1.Split(' ', ',').LastOrDefault();
+                string? line2LastWord = line2.Split(' ', ',').LastOrDefault();
+
+                if (string.Equals(line1LastWord, line2LastWord, StringComparison.CurrentCulture))
+                {
+                    return 0;
+                }
+
+                var newList = new List<string> { line1LastWord, line2LastWord }.OrderBy(word => word).ToList();
```
celluj34, line 710: This feels weird, can we use `string.Compare` directly instead? We should be able to replace these two entire lines with something like `string.Compare(line1LastWord, line2LastWord, StringComparison.CurrentCulture) == 0 ? -1 : 1`

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
```diff
@@ -626,6 +634,182 @@ private void ExecuteInverseCaseCommand()
...
+            {
+                string? line1LastWord = line1.Split(' ', ',').LastOrDefault();
+                string? line2LastWord = line2.Split(' ', ',').LastOrDefault();
+
+                if (string.Equals(line1LastWord, line2LastWord, StringComparison.CurrentCulture))
+                {
+                    return 0;
+                }
+
+                var newList = new List<string> { line1LastWord, line2LastWord }.OrderByDescending(word => word, StringComparer.CurrentCulture).ToList();
```
celluj34, line 748: Similar for this one, we could use `string.Compare` to avoid allocating a couple lists.

celluj34: Actually, we could replace everything after the `string.equals` on line 743 with `string.compare` and multiply by -1 (reversing the sort). It'll handle the 0 case automatically.

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
```diff
@@ -626,6 +634,182 @@ private void ExecuteInverseCaseCommand()
...
         #endregion
 
+        #region AlphabetizeCommand
+
+        public IRelayCommand AlphabetizeCommand { get; }
+
+        private void ExecuteAlphabetizeCommand()
+        {
+            string? text = _textBackup;
+            if (text is null)
```
veler, line 644: Please use `string.IsNullOrEmpty(text)` instead, since `"" != null`.

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
```diff
@@ -626,6 +634,182 @@ private void ExecuteInverseCaseCommand()
...
+            {
+                return;
+            }
+
+            IOrderedEnumerable<string> lines = text.Split('\r').OrderBy(line => line, StringComparer.CurrentCulture);
+            text = string.Join('\r', lines);
+
+            lock (_lockObject)
```
veler, line 652: We do this block of code at many places, including in the older code I wrote originally (oops). Could we create a helper method `SetTextWithoutBackup` that ultimately runs this block?

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
celluj34: Other commands only check `is null`, though?

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
```diff
@@ -626,6 +634,182 @@ private void ExecuteInverseCaseCommand()
...
+
+        private void ExecuteReverseAlphabetizeCommand()
+        {
+            string? text = _textBackup;
+            if (text is null)
+            {
+                return;
+            }
+
+            IOrderedEnumerable<string> lines = text.Split('\r').OrderByDescending(line => line, StringComparer.CurrentCulture);
```
veler, line 674: Could we use `Environment.NewLine` instead of `\r` ? It's more future-proof as DevToys 2.0 will target MacOS and Linux.

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
```diff
@@ -626,6 +634,182 @@ private void ExecuteInverseCaseCommand()
...
+            string? text = _textBackup;
+            if (text is null)
+            {
+                return;
+            }
+
+            var lines = text.Split('\r').ToList();
+            lines.Sort((line1, line2) =>
+            {
+                string? line1LastWord = line1.Split(' ', ',').LastOrDefault();
```
veler, line 740: `Split` method allocates N string (and so memory). I feel like this detection could be done without allocating extra strings (except the one we're interested in).

How about reading the string from its end to its beginning and detect word delimiter characters (any non-alphanumeric character?) ?

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
veler: Good point! Can't remember much why 😅
I will close this comment for now.

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
micahmo: Good idea! d423a5c86ac316c6f43518795d46217f51b1283b

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
micahmo: d423a5c86ac316c6f43518795d46217f51b1283b

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
micahmo: To add on to what @celluj34 said, I only checked against `null` for these reasons.

1. The existing commands only check against null.
2. Checking with `string.IsNullOrEmpty` is apparently not enough to tell the compiler that `text` is not null, forcing me to use a nasty null suppressor on subsequent lines.
   image
3. The sorting logic works fine with empty strings.

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
micahmo: You got it! 5f7d9b2ce6596e2680a6fe9a37dc9af09405ed09

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
micahmo: As mentioned in the original comment, the `CustomTextBox` was only giving me lines separated by `\r`, even when loading from a file with different line endings! `Environment.NewLine`, which is `\r\n` on Windows, would not work.

What I can do is split on all three variations (`\r`, `\n`, `\r\n`) to cover all cases. What do you think?

Or maybe you can check the behavior of DevToys 2.0 on another platform and see if `CustomTextBox` still has only `\r`.

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
micahmo: Sure, hopefully this does the trick: 429ffc252b67ac554fef2bdb546c0f41fdb97866

src/dev/impl/DevToys/ViewModels/Tools/Text/StringUtilities/StringUtilitiesToolViewModel.cs
veler: I see. Let's keep it as `\r` for now then. Will adapt this when migrating to DevToys 2.0 (FYI we will have to recode all the tools from scratch 😅)

veler (approved):
Approved. Thank you again for this improvement!

Issue #829, opened 2023-05-30
Human-readable description of cron expression


micahmo:
### What improvement do you think would an existing feature or tool in DevToys?

I really love the cron parser tool, thanks @DotNetInAction and @veler!

I was wondering if the output could contain a human-friendly description of the expression.

Here are some examples from [crontab.guru](https://crontab.guru/).
* `5 0 * 8 *`: "At 00:05 in August."
* `15 14 1 * *`: "At 14:15 on day-of-month 1."
* `0 22 * * 1-5`: "At 22:00 on every day-of-week from Monday through Friday."

### Solution/Idea

I find that having a human-readable description of the cron expression helps me to understand what it means, more than the "next" occurrence (although that's equally useful!).

### Comments

Possible help from the [cron-expression-descriptor](https://github.com/bradymholt/cron-expression-descriptor) library.

micahmo:
I am currently working on this using [cron-expression-descriptor](https://github.com/bradymholt/cron-expression-descriptor). It's working well, but I hit a snag with their library which I'm trying to fix over there: bradymholt/cron-expression-descriptor#164

Here's a sneak preview:

| ![JINvjlBVrM](https://github.com/veler/DevToys/assets/7417301/d851bebb-a830-4d66-bddb-732c0354673b) |
| - |

DotNetInAction:
Oh wow this looks cool! So glad to see people are actually using it!

micahmo:
#839

PR #839, opened 2023-06-05
Show description for cron expressions

micahmo:
## Pull request type

- [ ] Bugfix
- [ ] Build related changes

Issue Number: #829

## What is the new behavior?

This PR introduces a human-readable description of cron expressions to help users understand what they mean (in addition to seeing the evaluated times).

https://github.com/veler/DevToys/assets/7417301/6502af69-7f7a-41ba-a41f-9c04601630c3

### New PostBuildEvents
The CronExpressionDescriptor package comes with `.resource.dll` files for multiple cultures. By default, a folder containing each culture's resource will be copied to the output

image

However, UWP recompiles these into its own `resources.pri` file ([source](https://github.com/Humanizr/Humanizer/issues/317#issuecomment-66780033)). I didn't want to clutter the output with the unnecessary DLLs, so I added post-build events to the project files to clean these up. I tested multiple cultures, and it is able to correctly generate cron expressions for the current culture. If this is overkill and we don't care about these folders, I can revert the project file changes.

> P.S. I first tried to fix this the "right" way by explicitly referencing the undesired DLLs and setting CopyLocal to false as per [this recommendation](https://github.com/NuGet/Home/issues/5986#issuecomment-732401471), but that didn't work.

### Limitations

Cron Expression Descriptor doesn't support non-standard expressions like `@daily` or `@weekly`.

## Quality check

Before creating this PR:

   - [x] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)


src/dev/impl/DevToys/Views/Tools/Converters/CronParser/CronParserToolPage.xaml
```diff
@@ -102,7 +103,15 @@
              Tag="UserCronExpression"/>
 
         <controls:CustomTextBox
-             Grid.Row="2"
+            Grid.Row="2"
+            IsReadOnly="True"
+            AcceptsReturn="True"
```
veler, line 108: I think this should be false :)

veler:
Hello,

Thank you very much for your PR :)

> However, UWP recompiles these into its own resources.pri file (https://github.com/Humanizr/Humanizer/issues/317#issuecomment-66780033). I didn't want to clutter the output with the unnecessary DLLs, so I added post-build events to the project files to clean these up. I tested multiple cultures, and it is able to correctly generate cron expressions for the current culture. If this is overkill and we don't care about these folders, I can revert the project file changes.

That's quite interesting and I admit I'm puzzled by the fact that without the culture folder, the app still localize correctly the text.  Is it possible that your post build event clears the folders from `bin` folder but doesn't clear them from the generated `.msix` file?

I will double check that, if you don't mind.

src/dev/impl/DevToys/Views/Tools/Converters/CronParser/CronParserToolPage.xaml
micahmo: I left this as `True` because it affects the UI. When it is `True`, the copy button is placed above the text box and aligns nicely with the rest of the page.

| image |
| - |

When it is `False`, the copy button is placed inline and doesn't look as nice.

| image |
| - |

Since it's a readonly textbox, I figured `AcceptsReturn` won't have any negative affect.

What do you think?

micahmo:
> I admit I'm puzzled by the fact that without the culture folder, the app still localize correctly the text.

UWP is very new to me so I am by no means an expert here. However, even the existing application (before the new package) does not have resource folders or DLLs, yet it is able to localize correctly. I believe it must load resources from `resources.pri`. You can use `makepri.exe` to decompile this file and see all the strings. If you remove that file from the output, the app fails to start.

> it possible that your post build event clears the folders from `bin` folder but doesn't clear them from the generated `.msix` file?

Yes please verify that since I do not know how to generate the `.msix`. Thanks!

src/dev/impl/DevToys/Views/Tools/Converters/CronParser/CronParserToolPage.xaml
veler: That's fair! Let's keep it True then :)

veler:
> > I admit I'm puzzled by the fact that without the culture folder, the app still localize correctly the text.
> 
> UWP is very new to me so I am by no means an expert here. However, even the existing application (before the new package) does not have resource folders or DLLs, yet it is able to localize correctly. I believe it must load resources from `resources.pri`. You can use `makepri.exe` to decompile this file and see all the strings. If you remove that file from the output, the app fails to start.
> 
> > it possible that your post build event clears the folders from `bin` folder but doesn't clear them from the generated `.msix` file?
> 
> Yes please verify that since I do not know how to generate the `.msix`. Thanks!

Alright, I gave it a try by creating an MSIX package in Release mode. With the current PR, the text in the field you added stays in English even if DevToys is in French. By **NOT** removing the language folders, the text follows the right language.

Therefore, please remove these `PostBuildEvent` 😉

micahmo:
> Alright, I gave it a try by creating an MSIX package in Release mode. With the current PR, the text in the field you added stays in English even if DevToys is in French. By **NOT** removing the language folders, the text follows the right language.
> 
> Therefore, please remove these `PostBuildEvent` 😉

Gladly! My only concern was cluttering the output, but if these folders/dlls are necessary for localization, of course we will keep them. Thank you for testing this! Reverted in 302c1af471e05368f26d69f38df295fb323506a6.

veler (approved):
PR approved! Thank you so much again for this contribution! Keep the good work. We love all your recent contributions.

Issue #670, opened 2022-09-22
JWT Decoder show timestamp as a DateTime tooltip


sergiomontesa:

jlevier:
I tried looking into this but I could not get the hover to show on the Monaco editor.  I see we have this: https://github.com/veler/DevToys/blob/main/src/dev/impl/DevToys.MonacoEditor/ts-helpermethods/otherScriptsToBeOrganized.ts#L11 and just as a preliminary test I hardcoded a call in `CodeEditorCore.html`:

```javascript
monaco.languages.registerHoverProvider("json", {
    provideHover: function (model, position) {
        Debug.log("5555555555");
        return {
            range: new monaco.Range(
                1,
                1,
                2,
                5
                // model.getLineCount(),
                // model.getLineMaxColumn(model.getLineCount())
            ),
            contents: [
                { value: "**SOURCE**" },
                { value: "test test 1234"},
            ],
        };
    },
});
```

I also added "json" as the language type explicitly via 
```javascript
opt["language"] = 'json';
```

and then checked if `monaco.languages.registerHoverProvider` exists via
```javascript
Debug.log(typeof monaco.languages.registerHoverProvider);
Debug.log(monaco.languages.registerHoverProvider.toString());
```

[...]

Issue #607, opened 2022-07-22
Explain claims of JWT token like jwt.ms


cnshenj:
### What's the Problem?

The JWT Token Decoder only decodes property raw values. It would be much more helpful if each claim is explained in detail.
For example, both DevToys and jwt.ms show this:
```json
"iat": 1658523177,
```
But jwt.ms shows more if you switch to "Claims" tab:
<html>
<body>

iat | Fri Jul 22 2022 13:52:57 GMT-0700 (Pacific Daylight Time) | "Issued At" indicates when the authentication for this token occurred.
-- | -- | --

</body>
</html>

### Solution/Idea

Explain claims like jwt.ms.

### Alternatives

None.

### Priorities

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow devs to understand JWT token claims | Should |

### DevToys Version

Version 1.0.9.0 | X64 | RELEASE | 8bcfa60 | 8bcfa60

Christian-Oleson:
I am not sure how to see the "claims tab" but if someone wants to assign this one to me I am glad to take a look.

veler:
@btiteux , didn't you address that work your recent change? Should we close this issue then?

btiteux:
Hello, no I didn't resolve this issue but we discussed about it and how to do it.

I was going to take it but if @Christian-Oleson you want to take it go ahead.

We were thinking about adding a question icon or like a more information icon and we the user click this icon he will have a popin with the claims informations.

Christian-Oleson:
> Hello, no I didn't resolve this issue but we discussed about it and how to do it.
> 
> I was going to take it but if @Christian-Oleson you want to take it go ahead.
> 
> We were thinking about adding a question icon or like a more information icon and we the user click this icon he will have a popin with the claims informations.

A few thoughts then:
Here was something I did quickly, which only renders the found fields. If a field in the payload were sub, it would also include that information, for example:
![image](https://user-images.githubusercontent.com/54334220/190949203-05aa834c-9f33-430e-93b9-ca7cfcecc1ae.png)
This may not be what y'all want, and if so, that is fine. Just wanted to throw that out since I was getting familiar with the code.
As if you wish it to be an icon. Do we always want to show this icon? Or perhaps only when one of those payload fields is found? Also, where are you thinking about having it?

cnshenj:
> > Hello, no I didn't resolve this issue but we discussed about it and how to do it.
[...]

PR #847, opened 2023-06-08
Show dates on hover in JWT decoder

micahmo:
## Pull request type

- [ ] Bugfix
- [ ] Build related changes

Issue Number: #670

## What is the new behavior?

This PR adds the ability to show understandable dates in a tooltip when hovering over a timestamp in a decoded JWT.

As mentioned in the bug, my JavaScript and understanding of the architecture is pretty weak, so feel free to reject this if it doesn't fit the current model.

## Other information

https://github.com/micahmo/DevToys/assets/7417301/47bd90a2-0c90-4d73-b985-d19de06ee52b

## Quality check

Before creating this PR:

   - [x] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)

### DIFF
--- a/src/dev/impl/DevToys.MonacoEditor/CodeEditorControl/CodeEditorCore.cs
+++ b/src/dev/impl/DevToys.MonacoEditor/CodeEditorControl/CodeEditorCore.cs
@@ -1069,6 +1069,11 @@ namespace DevToys.MonacoEditor.CodeEditorControl
             return SendScriptAsync("editor.setPosition(" + JsonConvert.SerializeObject(position) + ");").AsAsyncAction();
         }
 
+        public void RegisterDateHoverProvider()
+        {
+            _ = InvokeScriptAsync("registerDateHoverProvider", "json");
+        }
+
         /// <summary>
         /// https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.icommoncodeeditor.html#deltadecorations
         /// 
--- a/src/dev/impl/DevToys.MonacoEditor/ts-helpermethods/otherScriptsToBeOrganized.ts
+++ b/src/dev/impl/DevToys.MonacoEditor/ts-helpermethods/otherScriptsToBeOrganized.ts
@@ -20,6 +20,32 @@ var registerHoverProvider = function (languageId: string) {
     });
 }
 
+var registerDateHoverProvider = function (languageId: string) {
+    return monaco.languages.registerHoverProvider(languageId, {
+        provideHover: function (model, position) {
+            var wordAtPosition = model.getWordAtPosition(position);
+            if (wordAtPosition === null) {
+                return null;
+            }
+            var date = new Date(Number(wordAtPosition.word) * 1000);
+            if (isNaN(date.getTime())) {
+                return null;
+            }
+            return {
+                range: new monaco.Range(
+                    position.lineNumber,
+                    wordAtPosition.startColumn,
+                    position.lineNumber,
+                    wordAtPosition.endColumn
+                ),
+                contents: [
+                    { value: date.toString() },
+                ],
+            };
+        }
+    });
+}
+
 var addAction = function (action: monaco.editor.IActionDescriptor) {
     action.run = function (ed) {
         Parent.callAction("Action" + action.id)
--- a/src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml.cs
+++ b/src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml.cs
@@ -189,6 +189,19 @@ namespace DevToys.UI.Controls
             set => SetValue(InlineDiffViewModeProperty, value);
         }
 
+        public static readonly DependencyProperty ShowDatesHoverProperty
+            = DependencyProperty.Register(
+                nameof(ShowDatesHover),
+                typeof(bool),
+                typeof(CodeEditor),
+                new PropertyMetadata(false));
+
+        public bool ShowDatesHover
+        {
+            get => (bool)GetValue(ShowDatesHoverProperty);
+            set => SetValue(ShowDatesHoverProperty, value);
+        }
+
         public CodeEditor()
         {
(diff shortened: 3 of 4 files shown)


btiteux:
Hello, Thank you for working on this request. 

I'm not really fan of the show on hover. I didn't add the human date format before, was looking for a way to display it without a hover maybe a small modal with a question mark to trigger it?

veler:
Hello,

Thanks @micahmo for this PR. I will join @btiteux on this. While I congratulate you for challenging the Monaco editor, I have 2 concerns about this tooltip approach:
1. There's no indication that passing the mice on the date will display any extra information. If I'm a new user and I'm not aware of it, I will likely never find it. In other words, it isn't very discoverable.
2. I'm also concerned by the accessibility aspect. This information is likely hard to get through a tooltip for someone who don't use a mice, or someone with vision impairment.

I would suggest to display this information in a permanent manner, either:
- By adding a comment in the JSON next to the timestamp (but I'm not 100% sure whether it can cause any issue. I admit I never used a JWT in production, @btiteux (or you?) may know).
- Add some text fields in the UI to display this info.

The comment in the JSON might be enough, and might be the best approach from a layout aspect (this tool already has a quite complex UI), but again, I'm not sure whether this could cause any frustration or not. I'm opened to opinions. :)

micahmo:
Hi @btiteux and @veler, thank you for the feedback and good conversation!

I will briefly push back on the notion that new users will not discover the tooltip. Since this is the design used by (what I believe is) the most popular JWT decoding site, [jwt.io](https://jwt.io/), I think people might actually expect this behavior in DevToys. 😉

https://github.com/veler/DevToys/assets/7417301/278c41c0-8771-476d-925d-cbeb53ac21db

Regardless, I totally understanding needing a different mechanism for accessibility and discoverability.

I believe I can condense both of your suggestions as follows:
 - A modal dialog
   - Personally, I am not a fan of dialogs to the extent they can be avoided.
 - Inline JSON comments
   - I did implement something like that [here](https://github.com/veler/DevToys/issues/670#issuecomment-1570657444). If we went with this approach, I would suggest a button to toggle the comments on/off so that the "clean" payload can be copied if needed. That said, this approach might not have the most longevity if we want to include more detailed descriptions of other claims as suggested in [#607](https://github.com/veler/DevToys/issues/607).
 - Additional UI fields
   - I like this idea the best, but I don't want the UI to become more cluttered. In 607, someone mentioned [jwt.ms](https://jwt.ms/), which is new to me. It has a tabbed interface where you can toggle between the raw decoded token and the claims fields explained. Although this site is brand new to me, I already like its approach the best. Instead of tabs, we could have an action button next to Copy that toggles the "explanation mode".

image

So my questions are:
 - Are you completely against the tooltip altogether, or could we keep it as a convenient, no-click solution for people who are expecting that behavior (in addition to another mechanism)?
 - For the additional mechanism, do you have a preference based on the ideas presented above?

Thanks again!

veler:
> A modal dialog
> Personally, I am not a fan of dialogs to the extent they can be avoided.
> Inline JSON comments
> I did implement something like that https://github.com/veler/DevToys/issues/670#issuecomment-1570657444. If we went with this approach, I would suggest a button to toggle the comments on/off so that the "clean" payload can be copied if needed. That said, this approach might not have the most longevity if we want to include more detailed descriptions of other claims as suggested in https://github.com/veler/DevToys/issues/607.
> Additional UI fields
> I like this idea the best, but I don't want the UI to become more cluttered. In 607, someone mentioned [jwt.ms](https://jwt.ms/), which is new to me. It has a tabbed interface where you can toggle between the raw decoded token and the claims fields explained. Although this site is brand new to me, I already like its approach the best. Instead of tabs, we could have an action button next to Copy that toggles the "explanation mode".

+1 to all of this! I completely agree with you.

> Are you completely against the tooltip altogether, or could we keep it as a convenient, no-click solution for people who are expecting that behavior (in addition to another mechanism)?

I think I will stay against for now. Even though I understand a good part of the population may know about it on jwt.io, I'm not fan of it. In addition, we are bleeding some extra options / info the `CodeEditor` control which will likely be only used by the JWT tool and therefore feel like CodeEditor will have an option specifically for 1 little scenario, which sounds a tiny unfair.

I'm willing to re-evaluate this opinion in the future, perhaps for DevToys 2.0.

> For the additional mechanism, do you have a preference based on the ideas presented above?

Yes! For DevToys 1.0 (and 2.0), I like the idea of an extra "Explanation" button. Although I'm not super enthusiastic about the name, but like the overall idea and can't think of anything else at the moment (maybe it will come to me later).

Regarding adding this "Explanation" button, how about the following mecanism:
To keep CodeEditor generic, we could add a ControlControl in the bar where the button appears. This ContentControl, tied to a new DependencyProperty, could display any kind of content (so not only button).

It could be use like this:

```xml
<CodeEditor>
    <CodeEditor.OptionBarAdditionalContent>
       <Button Content="Explanation" Click="OnClick"/>
    </CodeEditor.OptionBarAdditionalContent>
</CodeEditor>
```

This way, we could extend any other code editor in the future with extra options when it makes sense to do so.

What do you think?

micahmo:
Hey again, apologies for the delay on this! I tried a few different approaches and layouts, and I think this is the one I'd like to initially propose. Focusing just on the UI/UX, let me know what you think. If it's on the right path, I can clean things up and open a new PR. Thanks! I'm also happy to expoud on any of the design decisions (some of which were driven by technical limitations I hit).

https://github.com/veler/DevToys/assets/7417301/fa855d51-5627-45c2-9b1b-44b70c12b587

btiteux:
Hello, @micahmo I like the new Ui with the button to toggle more information (like the jwt.ms). And the more information when we click on the row is really nice.

What do you think @veler ?

veler:
I love it! :D

micahmo:
Thanks for the feedback! I will close this PR in favor of the rework in #864. I'll leave my branch for posterity in case it helps with tweaking Monaco in the future.

PR #848, opened 2023-06-08
Allow token validation without signature

micahmo:
## Pull request type

- [ ] Bugfix
- [ ] Build related changes

It is impossible to "validate" various parts of a JWT without providing the signature / signing key.

## What is the new behavior?

This PR separates signing key validation as a separate step from overall token validation. As a result, I can choose to validate only the expiration, for example, without providing a key, which is much more convenient.

https://github.com/veler/DevToys/assets/7417301/e3fd3a9d-a04f-4c53-ad62-4532e02853ec

## Other information

If we were starting from scratch, I think it would make sense for the overall validation operation to be called something like `ValidateToken`, and then it could have an option like `ValidateSignature`. However, the latter term is already used for the whole operation (and refers to a persistent setting), so I didn't want to change that. Instead I introduded a new option, `ValidateIssuerSigningKey` (which does align with the `TokenValidationParameters`), and as a result, `ValidateSignature` now has a slightly different meaning. I did update the UI strings to reflect this. Hopefully it's ok!

## Quality check

Before creating this PR:

   - [x] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)

### DIFF
--- a/src/dev/impl/DevToys/ViewModels/Tools/EncodersDecoders/JwtDecoderEncoder/JwtDecoder.cs
+++ b/src/dev/impl/DevToys/ViewModels/Tools/EncodersDecoders/JwtDecoderEncoder/JwtDecoder.cs
@@ -78,17 +78,26 @@ namespace DevToys.ViewModels.Tools.EncodersDecoders.JwtDecoderEncoder
             TokenParameters tokenParameters,
             TokenResult tokenResult)
         {
-            SigningCredentials? signingCredentials = GetValidationCredentials(tokenParameters);
             var validationParameters = new TokenValidationParameters
             {
-                ValidateIssuerSigningKey = true,
-                IssuerSigningKey = signingCredentials.Key,
-                TryAllIssuerSigningKeys = true,
                 ValidateActor = decodeParameters.ValidateActor,
                 ValidateLifetime = decodeParameters.ValidateLifetime,
                 ValidateIssuer = decodeParameters.ValidateIssuer,
                 ValidateAudience = decodeParameters.ValidateAudience
             };
+            
+            if (decodeParameters.ValidateIssuerSigningKey)
+            {
+                SigningCredentials? signingCredentials = GetValidationCredentials(tokenParameters);
+                validationParameters.ValidateIssuerSigningKey = decodeParameters.ValidateIssuerSigningKey;
+                validationParameters.IssuerSigningKey = signingCredentials.Key;
+                validationParameters.TryAllIssuerSigningKeys = true;
+            }
+            else
+            {
+                // Create a custom signature validator that does nothing so it always passes in this mode
+                validationParameters.SignatureValidator = (token, _) => new JwtSecurityToken(token);
+            }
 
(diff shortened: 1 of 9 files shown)


micahmo:
@veler Please tell me if I'm creating too much burden with my recnet PRs! I know you're pushing towards 2.0 and a lot of this will have to be rewritten anyway.

btiteux (commented):
Hello, @micahmo thank you for this improvement, no worries for the pull requests.  
The changes are good for me.

btiteux (approved):

src/dev/impl/DevToys/ViewModels/Tools/EncodersDecoders/JwtDecoderEncoder/JwtDecoder.cs
```diff
@@ -78,17 +78,26 @@ public class JwtDecoder
...
                 ValidateLifetime = decodeParameters.ValidateLifetime,
                 ValidateIssuer = decodeParameters.ValidateIssuer,
                 ValidateAudience = decodeParameters.ValidateAudience
             };
+            
+            if (decodeParameters.ValidateIssuerSigningKey)
+            {
+                SigningCredentials? signingCredentials = GetValidationCredentials(tokenParameters);
+                validationParameters.ValidateIssuerSigningKey = decodeParameters.ValidateIssuerSigningKey;
+                validationParameters.IssuerSigningKey = signingCredentials.Key;
```
veler, line 93: If `signingCredentials` is null, this will crash.

veler:
Would it be possible to add unit tests for the changes in this PR? :)

src/dev/impl/DevToys/ViewModels/Tools/EncodersDecoders/JwtDecoderEncoder/JwtDecoder.cs
micahmo: I believe this is ok for two reasons.
1. The same could be said for old code too, so I have not introduced a new problem.
2. It won't "crash"; instead, the exception will get caught and propagated to the UI. In this case, it will tell the user that the key is null, so they must enter a key if they want to validate the token. Again, that is the same as the existing behavior.

image

(Technically the error is from `GetValidationCredentials`, which should either throw or return a non-nullable. So I guess we could remove the nullable here to alleviate your concern. But again I was mostly refactoring the existing code.)

micahmo:
> Would it be possible to add unit tests for the changes in this PR? :)

Of course! d1740e4ddfcbc5598c3cedfdd8c190780cbebcbd

veler:
Awesome! Thank you so much ! :D
Looks like we have a small merge conflict now 😅 Should be good to go once resolved.

micahmo:
> Looks like we have a small merge conflict now 😅 Should be good to go once resolved.

Should be all set now! Thanks!

veler:
Thank you so much! :D

veler (approved):

PR #849, opened 2023-06-09
Fix JWT key validation

micahmo:
## Pull request type

- [x] Bugfix
- [ ] Build related changes

Unless I'm missing something obvious, it's currently impossible to validate tokens with a public key due to two issues:
 - The token algorithm is never set, so the key-building code always tries the wrong algorithm.
 - The public key text box is never displayed, because the visibility is only changed after a successful validation, which can't happen without the public key (chicken-and-egg).

(The unit tests have worked because they set the algorithm and public key explicitly.)

I also fixed a small issue related to **en**coding where the error banner would show when there wasn't an error and hide when there was!

## What is the new behavior?

**Before**: Don't even see the option for a public key. Obviously pasting it in the signature field doesn't work.

https://github.com/veler/DevToys/assets/7417301/4cc472c0-f78c-46fd-9703-0f86886ade93

**After**: Pasting a JWT with an RSA-encoded signature shows the Public key field and successfully validates.

https://github.com/veler/DevToys/assets/7417301/d1a93624-b1ef-4f82-9aa4-3e1b47d4d096

## Quality check

Before creating this PR:

   - [x] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)

### DIFF
--- a/src/dev/impl/DevToys/ViewModels/Tools/EncodersDecoders/JwtDecoderEncoder/JwtDecoder.cs
+++ b/src/dev/impl/DevToys/ViewModels/Tools/EncodersDecoders/JwtDecoderEncoder/JwtDecoder.cs
@@ -33,12 +33,13 @@ namespace DevToys.ViewModels.Tools.EncodersDecoders.JwtDecoderEncoder
         public TokenResult? DecodeToken(
                 DecoderParameters decodeParameters,
                 TokenParameters tokenParameters,
-                Action<TokenResultErrorEventArgs> decodingErrorCallBack)
+                Action<TokenResultErrorEventArgs> decodingErrorCallBack, out JwtAlgorithm? jwtAlgorithm)
         {
             Arguments.NotNull(decodeParameters, nameof(decodeParameters));
             Arguments.NotNull(tokenParameters, nameof(tokenParameters));
             _decodingErrorCallBack = Arguments.NotNull(decodingErrorCallBack, nameof(decodingErrorCallBack));
             Arguments.NotNullOrWhiteSpace(tokenParameters.Token, nameof(tokenParameters.Token));
+            jwtAlgorithm = null;
 
             var tokenResult = new TokenResult();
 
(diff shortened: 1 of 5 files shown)


btiteux (approved):
Hello, thank you for this fix, I totally forgot the public key validation.

src/dev/impl/DevToys/ViewModels/Tools/EncodersDecoders/JwtDecoderEncoder/JwtDecoder.cs
```diff
@@ -33,12 +33,13 @@ public class JwtDecoder
         public TokenResult? DecodeToken(
                 DecoderParameters decodeParameters,
                 TokenParameters tokenParameters,
-                Action<TokenResultErrorEventArgs> decodingErrorCallBack)
+                Action<TokenResultErrorEventArgs> decodingErrorCallBack, out JwtAlgorithm? jwtAlgorithm)
```
veler, line 36: Can you please bring this last parameter to another line? It helps readability.

veler:
I see you updated the unit tests, but I don't seem to find any added unit tests that covers this scenario. Can we please add some?

src/dev/impl/DevToys/ViewModels/Tools/EncodersDecoders/JwtDecoderEncoder/JwtDecoder.cs
micahmo: Done! 6d8dfd0c3a7a2068bbe1f0d1e2ced62dee02c7fc

micahmo:
> I see you updated the unit tests, but I don't seem to find any added unit tests that covers this scenario. Can we please add some?

Yes, I only updated the tests as needed to match the new signature of `DecodeToken`. But I agree, I should validate that the method properly detects and assigns the algorithm rather than specifying it in the test input. Done in 029433303d304d6ff738988b32e561816c90460e.

PR #859, opened 2023-06-13
JWT encoding/decoding improvements

micahmo:
## Pull request type

Please check the type of change your PR introduces:

- [x] Bugfix
- [x] New feature or enhancement
- [ ] UI change (please include screenshot!)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Internationalization and localization
- [ ] Other (please describe):

## What is the current behavior?

Issue Number: N/A

## What is the new behavior?

This PR includes a few improvements to JWT encoding/decoding.
 - Allow encoding tokens with RSA private keys
 - Make the header field readonly in encoding mode (since it doesn't currently affect the output, although that could be a future enhancement)
 - Allow validating signature with a certificate containing a public key
 - Fix an issue where the saved token algorithm was not loaded correctly

## Quality check

Before creating this PR:

- [x] Did you follow the code style guideline as described in [CONTRIBUTING.md](https://github.com/veler/DevToys/blob/main/CONTRIBUTING.md)
- [x] Did you build the app and test your changes?
- [ ] Did you check for accessibility? On Windows, you can use Accessibility Insights for this.
- [ ] Did you verify that the change work in Release build configuration
- [x] Did you verify that all unit tests pass
- [ ] If necessary and if possible, did you verify your changes on:
   - [x] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)

PR #864, opened 2023-06-23
Show additional JWT info

micahmo:
## Pull request type

- [ ] Bugfix
- [ ] Build related changes

Issue Numbers: #607 and #670

## What is the new behavior?

This PR adds the ability to show additional information about a decoded JWT. Key features are...
 - Each top-level property of the JWT is shown in a table with the key in one column and the value in another.
 - The values of date/time fields are shown in human-readable formats (instead of timestamp/long).
 - Clicking the row for a given field will display an explanation about that field (if it is a known/documented JWT claim).
   - There are descriptions for claims from [RFC 7519 Section 4.1](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1) and the [OpenID Connect Core Section 5.1, 2, 3.3.2.11, and 7.4](https://openid.net/specs/openid-connect-core-1_0.html) using the [IANA Registry](https://www.iana.org/assignments/jwt/jwt.xhtml#claims) as a source of truth.

## Other information

There are several things to mention about this implementation.
 - In order to inject the information, I went with a slightly different approach than was suggested. Rather than having the parent control inject the info button, it is part of the standard `CodeEditor`. However the info control itself can be set via a DependencyProperty, and the `CodeEditor` will show whatever's there. 
 - I struggled with how to disply the claim descriptions. They're usually very long, and I had trouble fitting them in a third column in the table. My options were (a) allow the grid to expand as wide as it wants, putting the info on one line and marking it hard to read, or (b) keep the grid width fixed and make the description wrap, which ended causing each row to consume too much vertical space. I thought about some other alternatives (truncating the description until clicked), but the bigger issue here is an apparent bug in the Toolkit DataGrid which causes rows taller than one line of text to have trouble rendering. I can upload a video of this if anyone's curious.
  Luckily, I ran across [Row Details](https://github.com/MicrosoftDocs/WindowsCommunityToolkitDocs/blob/main/docs/controls/datagrid_guidance/rowdetails.md) and I thought it would be the perfect compromise of having a nice clean grid initially with easy access to the descriptions. One quirk here is that I would expect clicking the row again to toggle the details off, but it doesn't. Maybe some special code can be added in the future to do that.
 - Along the same lines, I noticed that the grid doesn't respond perfectly to live theme changes. You have to scroll up and down to get all the rows to convert. I'm guessing this is another bug and maybe related to virtualization. However, I think it's ok since it's a rare case (changing theme while this control is open), and it eventually fixes itself.
 - Finally, there is some funny handling related to sizing. In addition to the technique used in the `CustomTextBox` (a `Border` whose size controls the `TextBox`'s size), I also had to add another trick. When expanding or collapsing the addiitonal info view, I quickly toggle on the `CodeEditor`. Without this, the `Border` has the wrong size. This does occasionally create a very small visual artifact, but it's the only way I could get the sizing to be solid.

https://github.com/veler/DevToys/assets/7417301/fa855d51-5627-45c2-9b1b-44b70c12b587

## Quality check

Before creating this PR:

   - [x] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)

### DIFF
--- a/src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml
+++ b/src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml
@@ -18,6 +18,8 @@
         <converters:InvertedBooleanConverter x:Key="InvertedBooleanConverter"/>
         <converters:BooleanToTextWrappingConverter x:Key="BooleanToTextWrappingConverter" TextWrappingOnTrue="Wrap" TextWrappingOnFalse="NoWrap"/>
         <converters:BooleanToIntegerConverter x:Key="ReadOnlyToClearButtonTabIndexConverter" ValueOnTrue="3" ValueOnFalse="0"/>
+        <converters:BooleanToVisibilityConverter x:Key="BooleanToVisibilityConverter"/>
+        <converters:BooleanToVisibilityConverter x:Key="InvertedBooleanToVisibilityConverter" IsInverted="True"/>
     </UserControl.Resources>
 
     <Grid>
@@ -124,7 +126,8 @@
             x:Name="CodeEditorCoreContainer"
             Grid.Row="1"
             Grid.Column="0"
-            Grid.ColumnSpan="2">
+            Grid.ColumnSpan="2"
+            Visibility="{x:Bind ShowInfoContent, Mode=OneWay, Converter={StaticResource InvertedBooleanToVisibilityConverter}}">
             <!--<monaco:CodeEditorCore
                 Grid.Row="1"
                 Grid.Column="0"
@@ -150,7 +153,7 @@
             Grid.Row="1"
             Grid.Column="0"
             Grid.ColumnSpan="2"
-            Visibility="Collapsed"
+            Visibility="{x:Bind ShowInfoContent, Mode=OneWay, Converter={StaticResource BooleanToVisibilityConverter}}"
             Content="{x:Bind InfoContent}">
         </ContentControl>
         <TextBlock
--- a/src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml
+++ b/src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml
@@ -61,6 +61,17 @@
                 <FontIcon x:Name="ExpandButtonIcon" Glyph="&#xF15F;"/>
             </Button>
             <Button
+                x:Name="InfoButton"
+                x:DeferLoadStrategy="Lazy"
+                Visibility="Collapsed"
+                TabIndex="0"
+                Margin="8,0,0,0"
+                ToolTipService.ToolTip="{Binding Instance.Common.ShowInfo, Mode=OneTime, Source={StaticResource LanguageManager}}"
+                AutomationProperties.Name="{Binding Instance.Common.ShowInfo, Mode=OneTime, Source={StaticResource LanguageManager}}"
+                Click="InfoButton_Click">
+                <FontIcon Glyph="&#xF4A3;"/>
+            </Button>
+            <Button
                 x:Name="PasteButton"
                 x:DeferLoadStrategy="Lazy"
                 Visibility="Collapsed"
@@ -126,6 +137,22 @@
                 DiffRightText="{x:Bind DiffRightText, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}"
                 AutomationProperties.LabeledBy="{x:Bind HeaderTextBlock}"/>-->
         </Grid>
+        <Border
+            x:Name="CodeEditorSizeFit"
+            Grid.Row="1"
+            Grid.Column="0"
+            Grid.ColumnSpan="2"
+            HorizontalAlignment="Stretch"
+            VerticalAlignment="Stretch"
+            SizeChanged="CodeEditorSizeFit_SizeChanged"/>
+        <ContentControl
+            x:Name="InfoGrid"
+            Grid.Row="1"
+            Grid.Column="0"
+            Grid.ColumnSpan="2"
+            Visibility="Collapsed"
+            Content="{x:Bind InfoContent}">
+        </ContentControl>
         <TextBlock
             Grid.Row="1"
             Grid.Column="0"
--- a/src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml.cs
+++ b/src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml.cs
@@ -219,7 +219,18 @@ namespace DevToys.UI.Controls
             set => SetValue(InfoContentProperty, value);
         }
 
-        private bool ShowInfoContent { get; set; }
+        public static readonly DependencyProperty ShowInfoContentProperty
+            = DependencyProperty.Register(
+                nameof(ShowInfoContent),
+                typeof(bool),
+                typeof(CodeEditor),
+                new PropertyMetadata(false));
+
+        public bool ShowInfoContent
+        {
+            get => (bool)GetValue(ShowInfoContentProperty);
+            set => SetValue(ShowInfoContentProperty, value);
+        }
 
         public CodeEditor()
         {
@@ -521,7 +532,6 @@ namespace DevToys.UI.Controls
             // when being expanded/collapsed. 
             bool showInfo = ShowInfoContent;
             ShowInfoContent = false;
-            UpdateInfoVisibility();
 
             IsExpanded = !IsExpanded;
             ExpandedChanged?.Invoke(this, EventArgs.Empty);
@@ -541,31 +551,13 @@ namespace DevToys.UI.Controls
             Window.Current.Dispatcher.RunIdleAsync(_ =>
             {
                 ShowInfoContent = showInfo;
-                UpdateInfoVisibility();
             });
         }
 
         private void InfoButton_Click(object sender, RoutedEventArgs e)
         {
             ShowInfoContent = !ShowInfoContent;
-
             ToolTipService.SetToolTip(GetInfoButton(), ShowInfoContent ? LanguageManager.Instance.Common.HideInfo : LanguageManager.Instance.Common.ShowInfo);
-
-            UpdateInfoVisibility();
-        }
-
-        private void UpdateInfoVisibility()
-        {
-            if (ShowInfoContent)
-            {
-                CodeEditorCoreContainer.Visibility = Visibility.Collapsed;
-                InfoGrid.Visibility = Visibility.Visible;
-            }
-            else
-            {
-                CodeEditorCoreContainer.Visibility = Visibility.Visible;
-                InfoGrid.Visibility = Visibility.Collapsed;
-            }
         }
 
         private async void PasteButton_Click(object sender, RoutedEventArgs e)
--- a/src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml.cs
+++ b/src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml.cs
@@ -206,6 +206,21 @@ namespace DevToys.UI.Controls
 
         public event EventHandler? ExpandedChanged;
 
+        public static readonly DependencyProperty InfoContentProperty
+            = DependencyProperty.Register(
+                nameof(InfoContent),
+                typeof(object),
+                typeof(CodeEditor),
+                new PropertyMetadata(null));
+
+        public object? InfoContent
+        {
+            get => GetValue(InfoContentProperty);
+            set => SetValue(InfoContentProperty, value);
+        }
+
+        private bool ShowInfoContent { get; set; }
+
         public CodeEditor()
         {
             SettingsProvider = MefComposer.Provider.Import<ISettingsProvider>();
@@ -300,6 +315,11 @@ namespace DevToys.UI.Controls
         private FontIcon GetExpandButtonIcon()
         {
             return (FontIcon)(ExpandButtonIcon ?? FindName(nameof(ExpandButtonIcon)));
+        }
+
+        private Button GetInfoButton()
+        {
+            return (Button)(InfoButton ?? FindName(nameof(InfoButton)));
         }
 
         private Button GetCopyButton()
@@ -488,10 +508,21 @@ namespace DevToys.UI.Controls
             {
                 GetExpandButton().Visibility = Visibility.Visible;
             }
+
+            if (InfoContent is not null)
+            {
+                GetInfoButton().Visibility = Visibility.Visible;
+            }
         }
 
         private void ExpandButton_Click(object _, RoutedEventArgs e)
         {
+            // This little trick is needed because the info content doesn't always size properly
+            // when being expanded/collapsed. 
+            bool showInfo = ShowInfoContent;
+            ShowInfoContent = false;
+            UpdateInfoVisibility();
+
             IsExpanded = !IsExpanded;
             ExpandedChanged?.Invoke(this, EventArgs.Empty);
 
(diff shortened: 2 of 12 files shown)


src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml.cs
```diff
@@ -505,6 +536,36 @@ private void ExpandButton_Click(object _, RoutedEventArgs e)
                 GetExpandButtonIcon().Glyph = "\uF15F";
                 ToolTipService.SetToolTip(GetExpandButton(), LanguageManager.Instance.Common.Expand);
             }
+
+            // Must dispatch this final resize to give the main control a chance to size properly
+            Window.Current.Dispatcher.RunIdleAsync(_ =>
```
veler, line 541: Is waiting for idle necessary? Perhaps https://github.com/veler/DevToys/blob/5b82435792870f193d8149e6d6b811e2d48a841c/src/dev/impl/DevToys/Core/Threading/ThreadHelper.cs#L37C30-L37C48 is enough, although I'd expect we already run on the UI thread here.

src/dev/impl/DevToys/ViewModels/Tools/EncodersDecoders/JwtDecoderEncoder/JwtClaim.cs
```diff
@@ -0,0 +1,50 @@
...
+            : _claim.Value;
+
+        public string? Description =>
+            TryGetDescription(_claim.Type.ToLowerInvariant(), out string? description)
+                ? description
+                : LanguageManager.Instance.JwtDecoderEncoder.NoDescription;
+
+        #region Statics
+
+        private static readonly ResourceLoader Resources = ResourceLoader.GetForViewIndependentUse(nameof(JwtDecoderEncoder));
```
veler, line 29: Please prefer using static property `LanguageManager.Instance.JwtDecoderEncoder` instead.

src/dev/impl/DevToys/Views/Tools/EncodersDecoders/JwtDecoderEncoder/JwtDecoderControl.xaml
```diff
@@ -186,7 +198,45 @@
...
             AllowExpand="True"
-            ExpandedChanged="PayloadCodeEditor_ExpandedChanged"/>
+            ExpandedChanged="PayloadCodeEditor_ExpandedChanged">
+            <controls:CodeEditor.InfoContent>
+                <Border
+                    BorderThickness="1" 
+                    BorderBrush="{ThemeResource TextControlBorderBrush}" 
+                    CornerRadius="{ThemeResource ControlCornerRadius}"
+                    Background="{ThemeResource TextControlBackground}">
+                    <toolkit:DataGrid
```
veler, line 208: @btiteux unfortunately this won't be fun to do in Blazor 😅 Hopefully we can reuse something open source to help us, or come up with an easier to implement view.

src/dev/impl/DevToys/Views/Tools/EncodersDecoders/JwtDecoderEncoder/JwtDecoderControl.xaml.cs
```diff
@@ -51,5 +65,22 @@ private void PayloadCodeEditor_ExpandedChanged(object sender, System.EventArgs e
...
+        }
+
+        private void ApplySettings()
+        {
+            if (_settingsProvider is not null && InfoDataGrid is not null)
+            {
+                FontFamily fontFamily = new(_settingsProvider.GetSetting(PredefinedSettings.TextEditorFont));
+                InfoDataGrid.FontFamily = fontFamily;
+                Style style = new(typeof(DataGridColumnHeader));
+                style.Setters.Add(new Setter(FontFamilyProperty, fontFamily));
```
veler, line 81: Can't we use Binding with a DependencyProperty instead?

src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml.cs
micahmo: > Is waiting for idle necessary?

Unfortunately I think so, because we need to wait until after the `CodeEditor`'s size has evaluated.

I tried `ThreadHelper.RunOnUIThreadAsync`, but as you said, we're already on the UI thread. I also tried `RunAsync` with every `CoreDispatcherPriority`, and none of them worked (actually `Idle` threw an exception).

I realize this is hacky, but it's the only way I could get the sizing to work. I can upload a video of the problem without this if you'd like.

src/dev/impl/DevToys/ViewModels/Tools/EncodersDecoders/JwtDecoderEncoder/JwtClaim.cs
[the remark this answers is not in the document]
micahmo: Sure! Done in 6197a6ba80e21075ad7873cfa44c0b07efa25d5f.

src/dev/impl/DevToys/Views/Tools/EncodersDecoders/JwtDecoderEncoder/JwtDecoderControl.xaml.cs
micahmo: I think I was able to do what you're suggesting in c1a64ccf75b936a72a8b8814d7819bc4d0611da4. I was also able to eliminate the `Loaded` handler, and I didn't need to set the column headers separately.

src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml.cs
veler: I see.
How about making `ShowInfoContent` a DependencyProperty and bind it to the Visibility property of `CodeEditorCoreContainer` and `InfoGrid` in the XAML by using a [BooleanToVisibilityConverter](https://github.com/veler/DevToys/blob/5b82435792870f193d8149e6d6b811e2d48a841c/src/dev/impl/DevToys/UI/Converters/BooleanToVisibilityConverter.cs#L12)?

I'm not sure to follow the sizing issue here otherwise.

micahmo: Good idea, binding the `Visibility` of both controls is a better approach and reduces the code-behind. I pushed this in d20981fb290ff6d3711cbd96cb744260015b0576.

However, it doesn't change the need to dispatch the property change. 😞

src/dev/impl/DevToys/Views/Tools/EncodersDecoders/JwtDecoderEncoder/JwtDecoderControl.xaml.cs
veler: It looks much better! :D

src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml.cs
veler: Thanks! Could we try one last thing to replace RunIdleAsync? How about:

```
ShowInfoContent = showInfo;
UpdateLayout();
```

This should trigger the UI to rearrange et remeasure itself, I believe.

src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml
```diff
@@ -126,6 +140,22 @@
                 DiffRightText="{x:Bind DiffRightText, Mode=TwoWay, UpdateSourceTrigger=PropertyChanged}"
                 AutomationProperties.LabeledBy="{x:Bind HeaderTextBlock}"/>-->
         </Grid>
+        <Border
```
veler, line 143: I think I start to understand better why you need the RunIdleAsync 🤔
If I follow correctly the logic, you have a Border that stretches so it fill-in its parent. Then you have this ContentControl that doesn't have the HorizontalAlignment and VerticalAlignment properties set to stretch. Therefore, to make it fitting, you listen to the SizeChanged event of the Border.

I admit I didn't give a try at the code yet, but seeing this make me wonder "why?". Why is this needed? Why not:
1. Setting VerticalAlignment and HorizontalAlignment to Stretch to the ContentControl?
2. Why not making the ContentControl a child of the Border?
```suggestion
        <Border
```

src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml.cs
micahmo: Unfortunately this does not work either. I believe the problem is that the CodeEditor (whose sizing we are using as the source of truth) never gets the chance to be drawn since we change `ShowInfoContent` before the UI thread runs.

I will continue the conversation in your other comment.

src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml
micahmo: I believe you are partly right, but let me clarify a few things.

> you have a Border that stretches so it fill-in its parent. Then you have this ContentControl that doesn't have the HorizontalAlignment and VerticalAlignment properties set to stretch. Therefore, to make it fitting, you listen to the SizeChanged event of the Border.

Yes, this is right. Actually, the Border technique is one that I stole from elsewhere in DevToys ([it is used for CustomTextBox](https://github.com/veler/DevToys/blame/main/src/dev/impl/DevToys/UI/Controls/CustomTextBox.xaml.cs#L856-L871)). I believe TextBox has the same problem as DataGrid which is that, without a fixed height, it will naturally use all of the vertical space, causing the parent page to scroll rather than the control itself. I think the Border technique is actually quite cool (by applying the sizing rules of one control to another).

> you need, sometime, to wait for the Border to get its size recalculated so you can give the ContentControl the correct size.

No, not quite. I think the Border technique should work fine (and does for CustomTextBux) But there is a problem which I cannot explain. When the DataGrid is displayed and the user collapses the expanded view, the Border sizing is wrong. It requests too much vertical space, despite not having any contents. This is where my second hacky trick comes into play. I briefly show the CodeEditor. Somehow, it sizes much better and fixes the Border. At that point I can put back the DataGrid, which is sized according to the Border.

Here is a video of the problem.

https://github.com/veler/DevToys/assets/7417301/9b437ffe-1c11-421e-9f6d-94a4284996d8


Both of your suggestions (ContentControl Alignment or child of Border) allow the DataGrid to consume as much vertical space as it wants, causing the parent tool page to scroll rather than the DataGrid itself.

P.S. Thank you so much for working with me on this. I think it's a really useful feature and I hope we can do it in a way that works and is nicely written. 😊

veler: hello,
Thanks for the explanation. That's fair then! :D Let's keep this solution. I'm just super surprised of the need of this `RunIdleAsync ` since I never needed it before 😅

micahmo: Sounds good, and trust me, I would love to have found a less hacky solution! The real problem is that we need the help of the `CodeEditor` to size the `DataGrid` when the `Border` doesn't work. `RunIdleAsync` is just the only way I could make the `CodeEditor` size after being made visible.

veler: That's fair 😁
I'm hoping we won't need anything hacky in DevToys 2.0 (made with Blazor, not Xaml) 🤞

veler:
Alright, I (finally) gave a try at the PR (so sorry for taking so long, focusing on DevToys 2.0 + my personal life on the side takes a lot of time).

It remains one "tiny" sizing issue on which I'm a little picky.

image

Here is the issue. and it sounds related to everything we mentioned previously:

https://github.com/veler/DevToys/assets/3747805/3cce650f-9e5d-4d3d-8b29-27cd87a8bcc1

I can also still reproduce the issue you demoed in the video in the previous comment, unfortunately. Do you reproduce it on your side too beside the `RunIdleAsync` ?

In the end, this PR is already super great and I don't think it's a blocker for merging it, but it would be amazing if we could address is before though (just to make it perfect). What do you think?

src/dev/impl/DevToys/UI/Controls/CodeEditor.xaml
micahmo: Sounds good. I'm a huge fan of XAML despite its quirks. I hope Blazor holds up! I might have to learn it. 😊

veler: I'm a huge Xaml fan too! Been doing it since 2006. For the first time, I meet a case where Xaml isn't the right answer. We can thank our dependency on a efficient and scalable text editor  along with cross-platform need. 😅

micahmo:
> so sorry for taking so long, focusing on DevToys 2.0 + my personal life on the side takes a lot of time

No worries whatsoever, I appreciate the tool and having the chance to contribute just a little. 😊 BTW, after this and #859 I would probably be done for a little while if you wanted to make a release (I'm not trying to dictate your release schedule or anything like that, just saying I added most of the things I wanted 😆).

> It remains one "tiny" sizing issue on which I'm a little picky.

The issue you demonstrated is actually unrelated to anything we've talked about so far. In this case, you are giving the `DataGrid` more room than it needs, and it doesn't seem to have any notion of expanding to be taller than the number of rows. Therefore, if the number of rows is very small, it will not occupy the total vertical space. This can be seen without resizing just by maximizing the window and providing a small JWT.

image

Let me see if there are any more hacky solutions I can come up with for this. 😆

veler:
Oh what I was referring to in fact is that the code editor on the left doesn't shrink at all as we make the window height smaller when the data grid is displayed, but does when the code editor on the right is displayed.

micahmo:
Ah I totally missed that. I was so focused on the right side! Yes, I can reproduce that too. Good catch! I will take a look.

micahmo:
@veler I've been meaning to ask you, when you run DevToys in debug mode, does it take a long time for the Monaco editor to load? For example, when I click on the JWT tool, it takes about 1.25 minutes while the editors spin until the UI is responsive. Is this normal or maybe I'm doing something wrong?

veler:
> @veler I've been meaning to ask you, when you run DevToys in debug mode, does it take a long time for the Monaco editor to load? For example, when I click on the JWT tool, it takes about 1.25 minutes while the editors spin until the UI is responsive. Is this normal or maybe I'm doing something wrong?

Not sure if "normal" but I always hit it too. There's a LOT happening to load Monaco due to the whole interop we need to do between the C# code and the JavaScript int he WebView. Even without debugger attached it takes 1-2 seconds to load (and we have loads of complain about it). My recommendation is that, when possible, start DevToys without Debugger, then attach the debugger through `Debug > Attach to Process` in Visual Studio.

Good news though: in DevToys 2.0, it takes ~50ms to load, whether you are with the debugger attached or not 🙌

micahmo:
Hi @veler, thanks for your patience on this. I have been working off and on for the past few days, but unfortunately, I can't get it to work as desired. The closest I could get was pulling the `DataGrid` into the parent `ToolPage` and then setting its height to some percentage less than 100% of the `ActualHeight` of the other `CodeEditor` (payload). Unfortunately, as soon as the heights match, the `DataGrid` takes more space, causing the `CodeEditor` to take more space, and they both keep expanding until the `DataGrid` is full height.

I believe we have the following options at this point.

1. Accept the PR as-is. It is not perfect, but maybe the new functionality will outweigh the minor UI issue.
2. Abandon the PR. I will take no offense at this as I totally understand that we want things to be perfect.
3. You or someone else can take a look. There's a good chance I missed something and we need a more expert eye.
4. Use an alternative control. Actually, my first idea was to format the data as an ASCII table (using a library) and put it in a `TextBox`, which might size better.

Again, I hate giving up, but I would rather move on than sink too much time into this one. 😊

veler:
> Hi @veler, thanks for your patience on this. I have been working off and on for the past few days, but unfortunately, I can't get it to work as desired. The closest I could get was pulling the `DataGrid` into the parent `ToolPage` and then setting its height to some percentage less than 100% of the `ActualHeight` of the other `CodeEditor` (payload). Unfortunately, as soon as the heights match, the `DataGrid` takes more space, causing the `CodeEditor` to take more space, and they both keep expanding until the `DataGrid` is full height.
> 
> I believe we have the following options at this point.
> 
> 1. Accept the PR as-is. It is not perfect, but maybe the new functionality will outweigh the minor UI issue.
> 2. Abandon the PR. I will take no offense at this as I totally understand that we want things to be perfect.
> 3. You or someone else can take a look. There's a good chance I missed something and we need a more expert eye.
> 4. Use an alternative control. Actually, my first idea was to format the data as an ASCII table (using a library) and put it in a `TextBox`, which might size better.
> 
> Again, I hate giving up, but I would rather move on than sink too much time into this one. 😊

No worries :D Thank you so much for trying hard on it. I took a brief look and I admit I don't see an easy solution either. 
I will approve and merge this PR. I think the benefits outweigh this detail. Worst case, we will get some issues opened and will act on it.

Thank you so much for this again! We love your contributions 😍

veler (approved):

PR #902, opened 2023-08-20
[DevToys 2.0] Added Base64 Text Encoder / Decoder tool

veler:
## Pull request type

- [ ] Bugfix
- [ ] Build related changes

DevToys 2.0 doesn't have Base64 Text Encoder / Decoder yet.

## What is the new behavior?

Ported the tool from DevToys 1.0

## Other information

image

image

image

## Quality check

Before creating this PR:

   - [X] Windows
   - [X] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)


src/app/dev/DevToys.Tools/Helpers/Base64Encoding.cs
btiteux, line 1: Hello, maybe we should create a models folder for this file no?

I was going to create one for the IndentationMode enum.

src/app/dev/DevToys.Tools/Tools/EncodersDecoders/Base64Text/Base64TextEncoderDecoderCommandLineTool.cs
```diff
@@ -0,0 +1,64 @@
...
+    [CommandLineOption(
+        Name = "encoding",
+        Alias = "e",
+        DescriptionResourceName = nameof(Base64TextEncoderDecoder.EncodingOptionDescription))]
+    private Base64Encoding EncodingMode { get; set; } = Base64Encoding.Utf8;
+
+    public ValueTask<int> InvokeAsync(ILogger logger, CancellationToken cancellationToken)
+    {
+        string output;
+        if (ConversionMode == Conversion.Encode)
```
btiteux, line 44: I think we should share this block of code between the Command Line and the GUI tool no?

src/app/dev/DevToys.Tools/Helpers/Base64Encoding.cs
veler: Good point ! Let's do this.

src/app/dev/DevToys.Tools/Tools/EncodersDecoders/Base64Text/Base64TextEncoderDecoderCommandLineTool.cs
veler: Good point!

src/app/dev/DevToys.Tools/Helpers/Base64Encoding.cs
btiteux: Hum just thinking about this, but that's depend if this Enum will be shared between different tools. 

If it's the case, then yes, I think the Models folder will be good, but if it's not the case we could put this file in the same folder as the tool no?

veler: Well, I don't mind trying to centralize things in various folder :)

src/app/dev/DevToys.Tools/Tools/EncodersDecoders/Base64Text/Base64TextEncoderDecoderCommandLineTool.cs
veler: Actually, nevermind. In the GUI tool, the parameter to figure out whether we Encode or Decode is a boolean. It's practical this way because the Switch() UI Element accepts a boolean for determining the state of the toggle switch.
Meanwhile, the enum is a better fit for Command Line because otherwise the user would have to type "true" or "false" when writing the parameters when invoking the DevToys.CLI app.

Let's keep it as it is then,

Issue #980, opened 2023-11-16
Lorem Ipsum: Generate other languages' Lorem.


RonStrauss:
### What improvement do you think would an existing feature or tool in DevToys?

A cool idea would be to generate dummy text in other languages/characters as well. 

### Solution/Idea

A `<select>` element with other languages, similar to what [this site](https://generator.lorem-ipsum.info/) offers.

PR #1004, opened 2023-12-17
Ported Lorem Ipsum generator to DevToys 2.0

veler:
## Pull request type

- [ ] Bugfix
- [ ] Build related changes

DevToys 2.0 doesn't have Lorem Ipsum generator yet.

## What is the new behavior?

Ported Lorem Ipsum generator. I also added an option that allows to use more corpus, as requested in #980.
In addition, instead of taking dependency on the library `NLipsum` like we do in DevToys 1.0, I forked the code. The reason why is that this library isn't maintained anymore and uses .NET Standard 1.0, which would requires DevToys 2.0 to ship with many more DLLs.

## Other information

image

## Quality check

Before creating this PR:

   - [X] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)


src/app/dev/DevToys.Tools/Helpers/LoremIpsum/LipsumGenerator.cs
```diff
@@ -0,0 +1,279 @@
...
+            data = GenerateWords(count);
+            options = Word.Medium;
+        }
+        else if (feature == Features.Characters)
+        {
+            data = GenerateCharacters(count);
+        }
+        else
+        {
+            throw new NotImplementedException("Sorry, this is not yet implemented.");
```
btiteux, line 88: Hello, maybe we should use resx string for this Exception message?

veler: The message is not displayed to the user. So let's keep it in English only. Therefore, hardcoded is fine. :-)

btiteux (approved):

PR #1009, opened 2023-12-21
Add DateConverter (old Timestamp Converter) to DevToys 2.0 

btiteux:
## Pull request type

- [x] Bugfix
- [ ] Build related changes

DateConverter / Timestamp Converter doesn't exist in DevToys 2.

Issue Number: N/A

## What is the new behavior?

Ported the Timestamp Converter from DevToys v1 to DevToys v2. At the same time, I added three new functionalities to the Timestamp Converter.

- Convert Date to Ticks
- Convert Date to Milliseconds
- Convert Date using custom Epoch

With these new features, the Timestamp Converter has been renamed Date Converter.

## Other information

image

## Quality check

Before creating this PR:

- [x] Did you check for accessibility? On Windows, you can use Accessibility Insights for this.
   - [x] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)


src/app/dev/DevToys.Api/Tool/GUI/Components/IUICard.cs
```diff
@@ -0,0 +1,89 @@
...
+    /// <param name="id">An optional unique identifier for this UI element.</param>
+    public static IUICard Card(string? id)
+    {
+        return new UICard(id);
+    }
+
+    /// <summary>
+    /// Sets the <see cref="IUISetting.InteractiveElement"/> of the setting.
+    /// </summary>
+    public static IUICard InteractiveElement(this IUICard element, IUIElement? uiElement)
```
veler, line 74: Did you mean to create this method? Looks like it should be removed, isn't it?

src/app/dev/DevToys.Api/Tool/GUI/Components/IUICard.cs
```diff
@@ -0,0 +1,89 @@
+﻿namespace DevToys.Api;
+
+/// <summary>
+/// A component that represents a setting, with a title, description, icon and <see cref="IUIElement"/> for the option value.
```
veler, line 4: Description needs to be updated.

src/app/dev/DevToys.Api/Tool/GUI/Components/IUICard.cs
```diff
@@ -0,0 +1,89 @@
...
+    public static IUICard InteractiveElement(this IUICard element, IUIElement? uiElement)
+    {
+        ((UISetting)element).InteractiveElement = uiElement;
+        return element;
+    }
+
+    /// <summary>
+    /// Sets the children to be displayed in the stack.
+    /// </summary>
+    public static IUICard WithChildren(this IUICard element, params IUIElement[] children)
```
veler, line 83: I do see that by default, if I put multiple items in this method, items are being stacked horizontally (despite the lack of Stack Panel, but CSS makes it happen this way):

image

In my opinion, this component should not assume to stack items vertically. Instead of accepting a list of children, I think it should accept a single child `IUIElement`, which can be anything from a button to a stack for example. That would be consistent with other similar containers that aren't explicitly made to support multiple items. For example, `UIDialog`, `IUIListItem`, `IUIGridCell`, `IUIDataGridCell`, `IUISetting.InteractiveElement` (there's an example in Settings tool where the InteractiveElement is a Stack), `IUIDataGrid.CommandBarExtraContent` (which is also a Stack in JSON > Table tool).

src/app/dev/DevToys.Api/Tool/GUI/Components/IUICard.cs
```diff
@@ -0,0 +1,89 @@
...
+        internal set => SetPropertyValue(ref _children, value, ChildrenChanged);
+    }
+
+    public event EventHandler? ChildrenChanged;
+}
+
+public static partial class GUI
+{
+    /// <summary>
+    /// A component that represents a setting, with a title, description, icon and <see cref="IUIElement"/> for the option value.
```
veler, line 55: Here too, and pretty much every other comments in this file :)

src/app/dev/DevToys.Api/Tool/GUI/Components/IUIDropDownListItem.cs
```diff
@@ -28,6 +28,35 @@ internal UIDropDownListItem(string? text, object? value)
...
+    public override bool Equals(object? other)
+        => Equals(other as UIDropDownListItem);
+
+    public bool Equals(UIDropDownListItem? other)
+    {
+        if (other is null)
+        {
+            return false;
+        }
+
```
veler, line 47: How about adding:
```suggestion
        if (ReferenceEquals(this, other))
        {
            return true;
        }
```
It basically check if these two objects are the exact same instance. It can help making the comparison slighlty faster if `Value` is a complex object to compare or if `Text` is a long string.

src/app/dev/DevToys.Api/Tool/GUI/Components/IUIDropDownListItem.cs
```diff
@@ -28,6 +28,35 @@ internal UIDropDownListItem(string? text, object? value)
...
+        {
+            return true;
+        }
+
+        return false;
+    }
+
+    public override int GetHashCode()
+    {
+        return HashCode.Combine(Text!.GetHashCode(), Value!.GetHashCode());
```
veler, line 58: I didn't know this method, that's pretty neat! 😁

src/app/dev/DevToys.Blazor/Components/UIElements/UICardPresenter.razor.scss
```diff
@@ -0,0 +1,7 @@
+.ui-card-presenter {
+    background-color: var(--ui-setting-group-presenter-inner-ui-setting-background-color);
+
```
veler, line 3: Is this background-color necessary? Card control already has a background color. Also, it uses a color specific to the Setting control, it worries me a bit.

src/app/dev/DevToys.Tools/Models/DateValueType.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace DevToys.Tools.Models;
+public enum DateValueType
```
veler, line 2: Internal?

src/app/dev/DevToys.Tools/Tools/Converters/Date/DateConverterCommandLineTool.cs
```diff
@@ -0,0 +1,162 @@
...
+        Name = "epoch",
+        Alias = "e",
+        DescriptionResourceName = nameof(DateConverter.EpochOptionDescription))]
+
+    internal string? EpochOption { get; set; }
+
+    [CommandLineOption(
+        Name = "timezone",
+        Alias = "tz",
+        IsRequired = true,
```
veler, line 38: Could we make this entry optional and detect the local timezone in `InvokeAsync` when `TimeZoneOption` is null?

Also, to understand how to use this parameter, I had to look at the code then read the documentation of `TimeZoneInfo.FindSystemTimeZoneById` to find what I should type to get it to work. This is quite difficult from a user standpoint.

The following code should help with getting the list of available entries:
```
            foreach (TimeZoneInfo z in TimeZoneInfo.GetSystemTimeZones())
            {
                Console.WriteLine(z.Id);
            }
```

I think this list may be localized and change based on the user language, which is another challenge.

Unfortunately, the `CommandLineOption` attribute doesn't allow us to inject some dynamically generated documentation, therefore, here are my suggestions:
1. Run the code above in a Console app and copy the result. Paste it in `DateConverter.TimezoneOptionDescription` as part of the explanation on how to use this parameter. In the comment of `DateConverter.TimezoneOptionDescription`, explain how you got this list. It will be important for translators that they run this code in order to get an accurate list.
2. Change `CommandLineOption` to support dynamically-generated documentation. This is likely a lot of work to do and I seriously discourage you to spend a lot of time on it.
3. If my ideas above aren't good enough or require too much more work, then let's create a GitHub issue for us to track, and please add a `// TODO` comment with a link to that issue.

src/app/dev/DevToys.Tools/Tools/Converters/Date/DateConverter.resx
```diff
@@ -0,0 +1,255 @@
...
+  </data>
+  <data name="Milliseconds" xml:space="preserve">
+    <value>Milliseconds</value>
+  </data>
+  <data name="DateOptionDescription" xml:space="preserve">
+    <value>Date to convert. Supported format: "ISO 8601"</value>
+    <comment>CLI parameter description</comment>
+  </data>
+  <data name="EpochOptionDescription" xml:space="preserve">
+    <value>Epoch to use. Supported format: "ISO 8601"</value>
```
veler, line 228: Could we provide an example of valid input?

src/app/dev/DevToys.Tools/Tools/Converters/Date/DateConverterCommandLineTool.cs
```diff
@@ -0,0 +1,162 @@
...
+        DescriptionResourceName = nameof(DateConverter.TimezoneOptionDescription))]
+    internal string? TimeZoneOption { get; set; }
+
+    [CommandLineOption(
+        Name = "format",
+        Alias = "f",
+        IsRequired = true,
```
veler, line 45: Since this option has a default, not need to make it required.
image

src/app/dev/DevToys.Tools/Tools/Converters/Date/DateConverterCommandLineTool.cs
```diff
@@ -0,0 +1,162 @@
...
+        Name = "number",
+        Alias = "n",
+        DescriptionResourceName = nameof(DateConverter.NumberOptionDescription))]
+    internal long? NumberOption { get; set; }
+
+    [CommandLineOption(
+        Name = "date",
+        Alias = "d",
+        DescriptionResourceName = nameof(DateConverter.DateOptionDescription))]
+    internal string? DateOption { get; set; }
```
veler, line 26: I am wondering if these two could be combined into one option of type `OneOf<long, string>` and make the option Required? Asking this because right now if I call `.\DevToys.CLI.exe dateconverter -tz test` I get no result (not even a message to tell that I'm missing an input, and the `tz` is invalid).

src/app/dev/DevToys.Tools/Tools/Converters/Date/DateConverterCommandLineTool.cs
```diff
@@ -0,0 +1,162 @@
...
+using DevToys.Tools.Helpers;
+using DevToys.Tools.Models;
+using Microsoft.Extensions.Logging;
+
+namespace DevToys.Tools.Tools.Converters.Date;
+
+[Export(typeof(ICommandLineTool))]
+[Name("DateConverter")]
+[CommandName(
+    Name = "dateConverter",
```
veler, line 11: how about `date` as an alias?

src/app/dev/DevToys.Tools/Tools/Converters/Date/DateConverter.resx
```diff
@@ -0,0 +1,255 @@
...
+    <value>Month</value>
+  </data>
+  <data name="OffsetTitle" xml:space="preserve">
+    <value>Offset</value>
+  </data>
+  <data name="SearchDisplayName" xml:space="preserve">
+    <value>Date Converter</value>
+  </data>
+  <data name="SearchKeywords" xml:space="preserve">
+    <value>Time Date Timezone Epoch</value>
```
veler, line 166: Timestamp?

src/app/dev/DevToys.Tools/Helpers/DateHelper.cs
```diff
@@ -0,0 +1,93 @@
...
+    private static readonly DateTimeOffset maximumZoneOffsetTimestamp = DateTimeOffset.MaxValue;
+
+    internal static bool IsValidTimestamp(long value)
+    {
+        if (value < minimumZoneOffsetTimestamp.Millisecond || value > maximumZoneOffsetTimestamp.Millisecond)
+            return false;
+        return true;
+    }
+
+    internal static async Task<ToolResult<DateTimeOffset>> ConvertToDateTimeUtcAsync(
```
veler, line 18: isn't it a bit overkill to make sure to go off the UI thread in this method? From what I'm seeing, there's nothing in this code that should take more than 1 millisecond. `SwitchOffMainThreadAsync` is likely the slowest work being done here actually.

src/app/dev/DevToys.Tools/Helpers/DateHelper.cs
```diff
@@ -0,0 +1,93 @@
...
+            DateFormat.Seconds => currentEpoch.AddSeconds(value),
+            DateFormat.Milliseconds => currentEpoch.AddMilliseconds(value),
+            _ => throw new NotSupportedException(""),
+        };
+
+        cancellationToken.ThrowIfCancellationRequested();
+        return new(conversionResult);
+    }
+
+    internal static async Task<ToolResult<long>> ConvertToLongAsync(
```
veler, line 38: Same here

src/app/dev/DevToys.Tools/Tools/Converters/Date/DateConverterGuiTool.cs
```diff
@@ -0,0 +1,1022 @@
...
+            );
+    }
+
+    private IUIDropDownListItem[] BuildTimeZoneItems()
+    {
+        ReadOnlyCollection<TimeZoneInfo> systemTimeZone = TimeZoneInfo.GetSystemTimeZones();
+        string timeZoneSelectedId = _settingsProvider.GetSetting(timeZoneIdSettings);
+        var timeZoneDropDownItems = new IUIDropDownListItem[systemTimeZone.Count];
+
+        if (!Regex.IsMatch(systemTimeZone.ElementAt(0).DisplayName, @"^\(UTC.*\).+$"))
```
veler, line 964: Do you know whether the UTC term may vary based on the customer's language?

src/app/dev/DevToys.Tools/Tools/Converters/Date/DateConverterGuiTool.cs
```diff
@@ -0,0 +1,1022 @@
...
+        {
+            // version < .Net6
+            // This implementation mitigates the changes in the strings
```
veler, line 967: This is super interesting and a very good catch! How did you find out there was this change between .NET versions?

src/app/dev/DevToys.Tools/Tools/Converters/Date/DateConverterGuiTool.cs
```diff
@@ -0,0 +1,1022 @@
...
+                        Cell(
+                            DateConvertDstGridRow.DateTime,
+                            DateConvertDstGridColumn.RightTitle,
+                            Stack()
+                                .Vertical()
+                                .SmallSpacing()
+                                .WithChildren(
+                                    Label()
+                                        .Style(UILabelStyle.Body)
+                                        .Text("UTC DateTime")
```
veler, line 940: Please place this text and the other above in a RESX.

src/app/dev/DevToys.Tools/Tools/Converters/Date/DateConverterGuiTool.cs
```diff
@@ -0,0 +1,1022 @@
...
+        {
+            _timeMillisecondsInputNumber.Editable();
+            return;
+        }
+        _timeMillisecondsInputNumber.ReadOnly();
+    }
+
+    private void OnCustomEpochChanged(bool useCustomEpoch)
+    {
+        _useCustomEpoch = useCustomEpoch;
```
veler, line 274: Could we store the result of `EpochStack` in a field, and `Disable()` / `Enable()` the stack based on whether `useCustomEpoch` is true or false?

src/app/tests/DevToys.UnitTests/Tools/Converters/DateConverterGuiToolTests.cs
```diff
@@ -0,0 +1,497 @@
...
+
+    [Theory(DisplayName = "Convert valid dateTime should return valid Ticks")]
+    [InlineData("1870-01-01T00:00:00.0000000+00:00", "UTC", 589799232000000000)]
+    [InlineData("1923-11-23T19:58:07.0000000+00:00", "UTC", 606806062870000000)]
+    [InlineData("1800-11-22T19:58:07.0000000+00:00", "UTC", 567990862870000000)]
+    [InlineData("1870-01-01T00:00:00.0000000-08:00", "Pacific Standard Time", 589799520000000582)]
+    [InlineData("1870-01-01T00:00:00.0000000+00:00", "Pacific Standard Time", 589799232000000000)]
+    [InlineData("1923-11-23T19:58:07.0000000Z", "Pacific Standard Time", 606806062870000000)]
+    [InlineData("1800-11-22T19:58:07.0000000Z", "Pacific Standard Time", 567990574870000000)]
+    public async Task ConvertValidDateTimeShouldReturnValidTicks(
```
veler, line 430: Some of these tests fail on my machine. I know you mentioned in private some race condition you're fighting, but I found this one failing even when I run it individually:

image

Not sure what's going on. Let me know if you'd like me to debug it on my own.

src/app/tests/DevToys.UnitTests/Tools/Converters/DateConverterGuiToolTests.cs
```diff
@@ -0,0 +1,497 @@
+﻿using System.Globalization;
```
veler, line 1: Thank you so much for adding so many tests ! 😁😁😁😁

src/app/tests/DevToys.UnitTests/Tools/Converters/DateConverterGuiToolTests.cs
btiteux: Yeah should be fixed in the next push

src/app/dev/DevToys.Tools/Tools/Converters/Date/DateConverterGuiTool.cs
btiteux: This portion of code is from the version 1

src/app/dev/DevToys.Tools/Tools/Converters/Date/DateConverterCommandLineTool.cs
btiteux: Yeah for the moment I added the ms doc link for timezone: https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/default-time-zones?view=windows-11&WT.mc_id=DT-MVP-5001664&source=post_page-----cff0e2b37f52--------------------------------#time-zones

But we nee to find a way to handle these case.

src/app/dev/DevToys.Blazor/Components/UIElements/UICardPresenter.razor.scss
btiteux: Yeah forgot to remove this color

src/app/dev/DevToys.Tools/Tools/Converters/Date/DateConverterGuiTool.cs
veler: I see! :D

src/app/tests/DevToys.UnitTests/Tools/Converters/DateConverterGuiToolTests.cs
veler: It works, thank you! :D

src/app/dev/DevToys.Api/Tool/GUI/Components/IUICard.cs
veler: Still need to remove it?

src/app/dev/DevToys.Api/Tool/GUI/Components/IUICard.cs
veler: Still need to update the description?

veler (approved):
It looks great, thank you very much! :D

PR #1015, opened 2023-12-30
Enabled Hover capability on Monaco Editor and fixed JSON serialization issue.

veler:
## Pull request type

- [X] Bugfix
- [ ] Build related changes

In preparation of @btiteux work on `JWT Encoder / Decoder`, I was curious to demonstrate how to use `Monaco Editor` API to show some text on a given span when the mouse pass the mouse hover it. Like this:

image

Unfortunately, I quickly faced a bug where the JSON we pass from .NET code to JavaScript code has some properties that aren't expected by Monaco.

To add a message on hover of a span in Monaco, we need to create an instance of [MarkdownString](https://github.com/veler/DevToys/blob/769a0e8fcbd91c85397d557c48f060bb17748fa7/src/app/dev/DevToys.Blazor/Components/Text/MonacoEditor/Monaco/MarkdownString.cs#L7-L18).
The minimum needed to make it work is to set a string to its `Value` property. However, that wasn't working.

The reason why is that when we send the `MarkdownString` to the JavaScript using `IJsRuntime.InvokeAsync`, we would get a JSON like this:

```json
{
    "value": "My text",
    "isTrusted": null,
    "supportThemeIcons": null,
    "supportHtml": null,
    "baseUri": null,
}
```

The problem is: in [these methods from Monaco](https://github.com/microsoft/vscode/blob/83b909c39f0ce5368d3a41a30c609de86a2e106e/src/vs/base/common/htmlContent.ts#L102-L121C2), it considers that this JSON isn't good if `supportThemeIcons` is null (instead of undefined, or with a boolean value).
As a consequence, the hover feature doesn't work.

## What is the new behavior?

As a fix, I created a method `PrepareJsInterop` that takes any object and convert it into a dynamic object, where null properties are excluded.

See the documentation on top of `PrepareJsInterop` to understand why I took this approach :-)

With this method invoked, the generated JSON is now:

```json
{
    "value": "My text"
}
```

## Other information

I considered some other approaches mentioned in https://github.com/dotnet/aspnetcore/issues/12685 but the best in my humble opinion was that `PrepareJsInterop` method.

In the future, once this PR is merged, creating a tooltip on hover in Monaco will be as simple as:

```csharp
TextModel model = await _monacoEditor.GetModelAsync();

if (model is not null)
{
    var decoration = new ModelDeltaDecoration
    {
        Range = new Monaco.Range
        {
            StartLineNumber = 1,
            StartColumn = 1,
            EndLineNumber = 1,
            EndColumn = 10
        },
        Options = new ModelDecorationOptions
        {
            HoverMessage
                = new[]
                {
                    new MarkdownString()
                    {
                        Value = "hello world !"
                    }
                }
        }
    };

    await _monacoEditor.ReplaceAllDecorationsByAsync(new [] { decorations });
}
```

## Quality check

Before creating this PR:

   - [X] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)

### DIFF
--- a/src/app/dev/DevToys.Blazor/Components/Text/MonacoEditor/MonacoEditor.razor.cs
+++ b/src/app/dev/DevToys.Blazor/Components/Text/MonacoEditor/MonacoEditor.razor.cs
@@ -77,13 +77,13 @@ public partial class MonacoEditor : RicherMonacoEditorBase
             options.MouseWheelZoom = false;
             options.OverviewRulerBorder = false;
             options.ScrollBeyondLastLine = false;
-            options.FontLigatures = false;
+            options.FontLigatures = true;
             options.SnippetSuggestions = "none";
             options.CodeLens = false;
             options.QuickSuggestions = new QuickSuggestionsOptions { Comments = false, Other = false, Strings = false };
             options.WordBasedSuggestions = false;
             options.Minimap = new EditorMinimapOptions { Enabled = false };
-            options.Hover = new EditorHoverOptions { Enabled = false };
+            options.Hover = new EditorHoverOptions { Enabled = true, Above = false };
             options.MatchBrackets = "always";
             options.BracketPairColorization = new BracketPairColorizationOptions { Enabled = true };
             options.RenderLineHighlightOnlyWhenFocus = true;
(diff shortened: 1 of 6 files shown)


src/app/dev/DevToys.Blazor/Components/Text/MonacoEditor/MonacoEditor.razor.cs
```diff
@@ -77,13 +77,13 @@ protected override async Task OnAfterRenderAsync(bool firstRender)
             options.MouseWheelZoom = false;
             options.OverviewRulerBorder = false;
             options.ScrollBeyondLastLine = false;
-            options.FontLigatures = false;
+            options.FontLigatures = true;
```
veler, line 80: Unrelated change, but nice to have. It enabled font ligatures as it says. Depending on user feedback, we could consider adding an option in `Settings` to enable/disable ligatures.

btiteux (approved):
Hello, everything seems ok for me. I may have feedback or updates during the JWT Decode & Encode dev. 

We will see at this moment.

Issue #749, opened 2023-01-25
JSON path test


Haingue:
Could you please add a window to test/find a JSON path ?
One bloc to import/copy a JSON content, one bloc to write a JSON path and one bloc to show the result.
Something like that: https://jsonpath.com/

almostSouji:
Was just about to suggest that after playing around with [insomnia](https://insomnia.rest/)'s jsonpath explorer (they use <https://github.com/JSONPath-Plus/JSONPath>) for a project. It's so useful! Being able to quickly stuff my json data into devtoys (potentially with parsing from raw js objects to proper json syntax, for added coolness points) and querying/exploring them with JSONpath would be amazing!

veler:
Added to DevToys 2.0

PR #1017, opened 2024-01-06
Added a JSONPath tester tool

veler:
## Pull request type

- [ ] Bugfix
- [ ] Build related changes

Issue Number: #749

## What is the new behavior?

Added a new tool dedicated to testing JSONPath.
Also, moved Regex Tester and XML Validator tools to a new `Testers` group.

## Other information

image

## Quality check

Before creating this PR:

   - [X] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)


src/app/dev/DevToys.Tools/Helpers/XmlHelper.cs
```diff
@@ -39,7 +39,7 @@ internal static bool IsValid(string? input, ILogger logger)
         }
         catch (Exception ex)
         {
-            logger.LogError(ex, "Invalid data detected 'input'", input);
+            logger.LogError(ex, "Invalid data detected.");
```
veler, line 42: Let's not log user input. Logging them can result in having the user data leaking. Logs can be retrieved by bots and upload them online.

src/app/dev/DevToys.Tools/Helpers/JsonHelper.cs
```diff
@@ -28,7 +28,8 @@ internal static bool IsValid(string? input, ILogger logger)
 
         try
         {
-            var jtoken = JToken.Parse(input);
+            using JsonReader reader = new JsonTextReader(new StringReader(input));
+            JToken? jtoken = await JToken.LoadAsync(reader, settings: null, cancellationToken);
```
veler, line 32: Changed this method so we can pass the CancellationToken of SmartDetection (which cancels within 2 seconds or if the clipboard content changed)

veler:
> Hum quick thinking, should we keep the XML Validators in the same group or we could create a new Validators groups?
> 
> I think it's weird to have the XML Validators with the Testers.

Thanks for the suggestion. `XML Validator`, in the end, "tests" an XML against its XSD. How about renaming XML Validator to `XML / XSD Tester` or `XML Tester` ? What do you think @btiteux ?

src/app/dev/DevToys.Tools/Helpers/JsonHelper.cs
btiteux: We could use a ValueTask here.

src/app/dev/DevToys.Tools/Helpers/XmlHelper.cs
btiteux: Totally agree on this.

btiteux:
Ok we can rename the XML Validator for XML Tester.

No test for this tool?

veler:
> Ok we can rename the XML Validator for XML Tester.

Done !

> No test for this tool?

There are tests for the helper, but I admit there aren't for the gui and command line.

btiteux (approved):
Ok I will approve this pull request but we really need to add Unit Test for the GUI

veler:
Thanks @btiteux , I just pushed a few more tests for UI and command line

Issue #468, opened 2022-03-28
Escape tool escapes new-lines as '\r'


ShlomiAltostra:
### Description

The escape tool escapes new-lines as `\r`s

### Steps To Reproduce

Go to 'Text Escape / Unescape'
Make sure it's in "Escape" mode
Insert inputs which include a few lines of text

The 'Output' now has `\r` instead of `\n`s (or `\r\n`, though preferably just `\n`) instead of line breaks

### Expected behavior

New llines should be escaped as `\n` a line-feed character instead of a carriage-return

### DevToys Version

1.0.5.0

### Relevant Assets/Logs

![image](https://user-images.githubusercontent.com/57314279/160353913-f5b9b4b3-9c9e-4cca-bb18-aa6d3883c766.png)

veler:
Hello,
Thanks for your feedback.
This actually makes me wonder: perhaps we need an option allowing to select whether the return carrier character should be `\n`, `\r` or `\r\n`. What do you think?

Ping @btiteux and @rlm96 too.

rlm96:
I think that's a good idea, that way we can just switch the carrier character.

AHeiming:
Hi,
if this is okay for you, it would like to try solving this, though it's the first time I'm working with .xaml files and a Microsoft Store App 😅 

I hope that's okay for you.

AHeiming:
![image](https://user-images.githubusercontent.com/44135066/176504903-5dc5e695-5c0a-4b50-86d8-72d601096818.png)

This is how I thought it should look.
Icon is just temporary, any suggestions on the icon?

veler:
Hey @AHeiming , thanks for your suggestion.
I like this idea. Here is a suggestion to make it (hopefully) even better:
1. Instead of using some Radio Button, use a Combo Box, like for the option in `JSON Formatter`.
2. Instead of just displaying `\r`, `\n`, `\r\n`, show `CRLF (\r\n) (Windows)`, `CR (\r) (Mac)` and `LF (\n) (Linux)`.

Quick question:
Is that something that currently work on your machine or is this just a mock up?
Please feel free to open a PR (even draft). We can iterate on it together 😊

Thank you !

AHeiming:
Hi,

thanks for your suggestions! This is currently working on my machine, but not finished yet.
I hope to get this done this weekend.

I have a question, why a Combo Box? Should it behaive like a Radio Button? Because a multi selection does not seem working here for me.

veler:
Hi,

[...]

PR #971, opened 2023-11-06
Escape tool escapes new lines as (#1)

heribertolugo:
## Text Escape/Unescape Newline Encoding Options

Please check the type of change your PR introduces:

- [ ] Bugfix
- [x] New feature or enhancement
- [x] UI change (please include screenshot!)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Internationalization and localization
- [ ] Other (please describe):

## What is the current behavior?

Currently escaping or unescaping newlines in text only handles \r
Issue Number: #468

## What is the new behavior?
Allows the user to choose encoding option for newlines between CRLF (\r\n) (Windows), CR (\r) (Mac) and LF (\n) (Linux)

- Adds a combox withe the following options in the Text Escape/Unescape UI: CRLF (\r\n) (Windows), CR (\r) (Mac) and LF (\n) (Linux)
- Replaces all newline charcters to the option selected when escaping
- Converts all escaped newline charcters to the option selected

## Other information

before
image

after
image

## Quality check

Before creating this PR:

- [x] Did you follow the code style guideline as described in [CONTRIBUTING.md](https://github.com/veler/DevToys/blob/main/CONTRIBUTING.md)
- [x] Did you build the app and test your changes?
- [ ] Did you check for accessibility? On Windows, you can use Accessibility Insights for this.
- [x] Did you verify that the change work in Release build configuration
- [x] Did you verify that all unit tests pass
- [x] If necessary and if possible, did you verify your changes on:
   - [x] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)

PR #1019, opened 2024-01-09
Ported Text Escape/ Unescape tool

veler:
## Pull request type

- [ ] Bugfix
- [ ] Build related changes

DevToys 2.0 doesn't have a Text Escape / Unescape tool yet

Issue Number: #468

## What is the new behavior?

Ported the tool from DevToys 1.0 to 2.0. Issue #468 is solved through the fact that DevToys 2.0 has an option to select what end of line character should be used in Settings. Let's start with this for now. If unhappy, we could create an option specifically in this tool.

## Other information

image

## Quality check

Before creating this PR:

   - [X] Windows
   - [ ] macOS (DevToys 2.0)
   - [ ] Linux (DevToys 2.0)


btiteux (approved):
I'm ok with the pull request but we need some Unit Tests on the GUI  :)

Before merging this pull request #971 , we have one to add support for Linux escape lines, too. Can we add the support at the same time?

veler:
Hey @btiteux 

> Before merging this pull request #971 , we have one to add support for Linux escape lines, too. Can we add the support at the same time?

As I said in the PR description, issue #468 (which #971 solves in DevToys 1.0) is solved through the fact that DevToys 2.0 has an option to select what end of line character should be used in Settings. I'd suggest to limit ourselves to this for now and wait for more feedback. The reason of this push back is that the end of line sequence is set at the Monaco editor level. So replacing end of line in the string we hold in memory won't be efficient because Monaco will overwrite it constantly (when the user do any edit, or when copying or pasting into the editor).

A workaround is to add a new API to DevToys.API that would allow to customize the end of line sequence for a given `IUIMultiLineTextInput`. I'd like to try to avoid it (for now) in order to keep the API simple and don't give too much freedom (which translates into a risk of inconsistency between tools).

The setting I'm referring to above has already been tested on Linux.

image

Note that `\r (Mac)` is not supported by Monaco. Note also that `\r` on Mac is not supported after Mac OS 9, which DevToys doesn't target anyway.

veler:
> we need some Unit Tests on the GUI :)

Done :-)

btiteux:
We can go with the settings for the moment. But I think it's more user friendly to have a list here and that would avoid changing the settings for the whole apps.

veler:
> We can go with the settings for the moment. But I think it's more user friendly to have a list here and that would avoid changing the settings for the whole apps.

I ultimately agree :-)

PR #1108, opened 2024-04-01
Add base64 signature option to jwt encoder/decoder tool & some fixes

### LINKED ISSUE #1031: Decode date and time for Unix time values in JWT Decoder
### What improvement do you think would an existing feature or tool in DevToys?

Allow dates to be decoded into real dates and times and displayed in a hover-over tool-tip inside the JWT Decoder.

### Solution/Idea

 * Decode tokens containing Unix time values into a full date time string and show it in a tool-tip.
 * Well known Unix time values such as `nbf`, `exp`, `iat` and `auth_time` must be converted by default.
 * There should be an option to infer Unix time for any integer value greater than 1,000,000,000 (2001-09-09).

### Comments

Maybe the exact number used for identifying a Unix time value should be configurable? The idea behind this default/suggested value is to have a large number that is less likely to show up in other, run-of-the-mill configuration data but could reasonably be expected to be a Unix time. It might also be worth allowing floating point values rather than just integers.

There could be an option to add the decoded time to a comment in the decoded JSON(C). e.g.:

```jsonc
{
  "nbf": 1705425338, // 2024-01-16 17:15:38
  "exp": 1705428938, // 2024-01-16 18:15:38
  ...
}
```

akarboush:
## Pull request type

- [X] Bugfix
- [X] Code style update (formatting, renaming)
- [ ] Build related changes

- It is not possible to specify whether the provided JWT signature is in base64 format or not; it is implicitly inferred, which is not always accurate and may be unclear to the user.
- When hovering over payload date values, nothing occurs.

## What is the new behavior?

https://github.com/DevToys-app/DevToys/assets/40485241/a9ef44c9-fab4-4a80-8eef-221b3c1ae547

https://github.com/DevToys-app/DevToys/assets/40485241/c99c5662-c516-4ef1-a12b-04255ec4b275

Probably fixes #1031 

image

## Other information

Alongside the new feature, many bugs also occurred during implementation which I tried to fix. The majority of the changes were in the tests and tests data. This is because the tests weren't verifying that the output remained consistent throughout the entire encode/decode process. For instance, one bug was setting 'IssuedAt' to 'DateTime.Now', which resulted in the generated token not matching the user's specifications.

The commits should clearly indicate which bugs they fix, but here is a brief summary:
- The tools were not showing any errors (ClearUI was called).
- The algorithm type was created with the algorithm signature rather than the actual name (http://www.w3.org/2001/04/xmldsig-more#hmac-sha256 instead of sha256).
- Incorrect setting title.
- The Issuer and the audience were not populated from settings but were used when encoded/decoded.
- As mentioned above, 'IssuedAt' was set to 'DateTime.Now'.
- Tests were refactored and checked against an already generated token.
- Set the header input of the encoder tool to read-only, as it is currently completely ignored. (Dev-Toys 1 was read-only)
- Some refactoring and memory improvements.

Finally, as you might have noticed, I believe there's a styling issue with the switch's position in the decoder tool, I would gladly fix it but tbh I'm not exactly sure how to :(.

Screenshot

## Quality check

Before creating this PR:

   - [X] Windows
   - [ ] macOS
   - [ ] Linux


akarboush:
Just noticed this commit https://github.com/DevToys-app/DevToys/commit/70a4ea14f039dd93162ae81254158a7bc4169458 and it looks like there is a new repo for the tools now. I should probably close this PR

Issue #1228, opened 2024-06-15
Windows 11 misaligned title bar color


veler:
### Current behavior

The color on title bar is misaligned. It should be invisible (non colorized):
![image](https://github.com/DevToys-app/DevToys/assets/3747805/e0a5b02c-b190-4106-b137-496bbd496ca9)

### Expected behavior

The colorized title bar should be smaller.

### Affected platforms

Windows

### Affected DevToys kind

DevToys (app with GUI)

### DevToys Version

2.0.1

veler:
I reproduce the same issue in WPF Gallery App interestingly.

![image](https://github.com/DevToys-app/DevToys/assets/3747805/0bd5c657-7688-4939-8a12-a8b52f307e90)

0x5bfa:
Fyi, DevHome seemed to had this issue.

https://raw.githubusercontent.com/microsoft/devhomegithubextension/main/build/store/canary/media/ja-JP/github-setup.png

![image](https://github.com/DevToys-app/DevToys/assets/62196528/da58e2f4-571a-4c66-a84c-48ed969f46dc)

Is this some sort of issue with WinUI?

niyari:
WIP

veler:
Devtoys window is in WPF, so my guess is that it's a wpf specific problem, as your screenshots point it.

@niyari , did you find anything yet? I'm really struggling with this one.

niyari:
I found that this was caused by the background color of the WPF content and WebView being set to transparent by default.
![03](https://github.com/DevToys-app/DevToys/assets/2970321/6b9b3581-88e9-48e0-a6af-5a8db32d2c37)

Specifying a background color in the Grid or webview tags in MainWindow.xaml or in the HTML content (e.g. body) in DevToys hides the WPF title bar, giving it a more natural appearance.
![image](https://github.com/DevToys-app/DevToys/assets/2970321/aa3168f9-769b-4e2e-8dd6-839a570a4a80)

The easiest fix is ​​to make the title bar color a constant using CSS.

PR #1286, opened 2024-07-02
Fix .nav-bar-header background color.

niyari:
## Pull request type

- [ ] Bugfix
- [ ] Build related changes

Issue Number: #1228 

## What is the new behavior?

- Title bar is non colorized.
image

## Other information

note: The splash screen is not affected by this change.

## Quality check

Before creating this PR:

   - [x] Windows
   - [ ] macOS
   - [ ] Linux


veler:
Hi @niyari ,
Thanks for this PR. I think it can cause a problem for users who have the `Show accent color on title bar and window borders` option turned OFF in Windows Settings. It makes the title bar not having the Mica effect, because the color you defined hides it.

The change you suggest also affect Linux and MacOS.

Here is a suggestion:

Instead of changing the CSS, let's change the [MainWindow.xaml](https://github.com/DevToys-app/DevToys/blob/ab92874a3b1bc1b46d74b31d8ecd96744b72b031/src/app/dev/platforms/desktop/DevToys.Windows/MainWindow.xaml#L22-L22) and add a `Border` behind the `CustomBlazorWebView`, placed at the top of the window.

This Border could conditionally have a color:
1. When `Show accent color on title bar and window borders` is OFF, the Border should have a transparent background
2. When ON, the Border should have the same color than the one defined in Windows Settings.

These links can be helpful:
- https://stackoverflow.com/a/69896088
- https://copilot.microsoft.com/sl/dPmg0e0Gq5Y
- https://blog.codeinside.eu/2016/04/26/get-the-windows-10-or-8-accent-color-from-wpf/

veler:
Hey @niyari ,

Thanks again for your change proposal. I made a fix here that follows my suggestion: https://github.com/DevToys-app/DevToys/pull/1299

niyari:
I think it would be nice to unify the app's colors, similar to Visual Studio.
However, I didn't have time to suggest this change.

Issue #1381, opened 2024-08-16
Linux: Use desktop portals to get dark theme information


badcel:
>> If I might chime in on this issue:
I noticed you changed the code to check the GTK theme name in order to check for dark mode.
This is not an ideal way to check it. In fact, on a modern Gnome Desktop, it doesn't detect dark mode at all:

![Bildschirmfoto vom 2024-08-16 09-11-12](https://github.com/user-attachments/assets/73ddb9a5-cfed-4e15-94c3-53d184a72e8a)

> To give a few details:
While setting dark mode with the GTK theme name was a popular workaround when there was no proper standard, it still has it flaws. For example, `GTK_THEME` is regarded for use in debugging, not to set a dark theme. And some frameworks, like libadwaita, even ignore it entirely. Which is also why Gnome is not setting the GTK theme name when setting dark mode.

> Instead, the proper way would be to check the [Settings Portal](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Settings.html), part of the XDG Desktop Portals. The dark mode preference of Gnome sets the value read by the portal. And the libadwaita code you've used read the dark mode preference from this portal.

> Since GTK does not include support for a dark mode preference, you would need to read the portal directly. You could do this by directly accessing the DBus interface with a library like [Tmds.Dbus](https://github.com/tmds/Tmds.DBus). There's also a C library called [libportal](https://libportal.org/), but I'm not sure what the easier way would be.

> TLDR: Using the GTK theme name to check for dark mode is not ideal (maybe for a fallback), ideally the Settings portal should be used.

_Originally posted by @CodedOre in https://github.com/DevToys-app/DevToys/issues/1378#issuecomment-2293034566_

badcel:
You don't need an additional DBus library as there is already a DBus implementation inside GLib which is part of Gir.Core.

[I wrote a small sample to detect the color-scheme which works for me on fedora 40](https://github.com/gircore/gir.core/blob/main/src/Samples/Gio-2.0/DBus/ReadDesktopAppereanceColorScheme.cs). Any feedback that it works on other current distributions is very welcome.

veler:
Hi @badcel ,

Thank you soooo much for taking the time to explain all this and to provide a sample. You made my life sooo much easier !

I made a pull request that followed your advices: https://github.com/DevToys-app/DevToys/pull/1382

Have a wonderful day !

PR #1382, opened 2024-08-17
Fixed a regression where Linux theme is not properly detected

veler:
## Pull request type

- [X] Bugfix
- [ ] Build related changes

Issue Number: #1381

## What is the new behavior?

Use DBus to retrieve the `org.freedesktop.appearance` parameter to detect the theme on Linux Mint, Ubuntu, Fedora.

## Quality check

Before creating this PR:

- [X] Did you check for accessibility? On Windows, you can use Accessibility Insights for this.
   - [ ] Windows
   - [ ] macOS
   - [X] Linux

### DIFF
--- a/src/app/dev/platforms/desktop/DevToys.Linux/Core/ThemeListener.cs
+++ b/src/app/dev/platforms/desktop/DevToys.Linux/Core/ThemeListener.cs
@@ -2,14 +2,18 @@
 using DevToys.Core.Settings;
 using DevToys.Blazor.Components;
 using DevToys.Blazor.Core.Services;
+using Gio;
+using GLib;
+using Microsoft.Extensions.Logging;
 using Object = GObject.Object;
 using static GObject.Object;
 
 namespace DevToys.Linux.Core;
 
 [Export(typeof(IThemeListener))]
-internal sealed class ThemeListener : IThemeListener
+internal sealed partial class ThemeListener : IThemeListener
 {
+    private readonly ILogger _logger;
     private readonly ISettingsProvider _settingsProvider;
     private readonly Gtk.Settings _gtkSettings;
 
@@ -19,6 +23,8 @@ internal sealed class ThemeListener : IThemeListener
     [ImportingConstructor]
     public ThemeListener(ISettingsProvider settingsProvider)
     {
+        _logger = this.Log();
+
         // Listen for app settings
         _settingsProvider = settingsProvider;
         _settingsProvider.SettingChanged += SettingsProvider_SettingChanged;
@@ -147,8 +153,44 @@ internal sealed class ThemeListener : IThemeListener
 
     private AvailableApplicationTheme GetCurrentSystemTheme()
     {
-        return _gtkSettings.GtkApplicationPreferDarkTheme || (_gtkSettings.GtkThemeName?.Contains("Dark", StringComparison.OrdinalIgnoreCase) ?? false)
-            ? AvailableApplicationTheme.Dark
-            : AvailableApplicationTheme.Light;
+        try
+        {
+            var bus = DBusConnection.Get(BusType.Session);
+            using var parameters = Variant.NewTuple([
+                Variant.NewString("org.freedesktop.appearance"), Variant.NewString("color-scheme")
+            ]);
+
+            using Variant ret = bus.CallSync(
+                busName: "org.freedesktop.portal.Desktop",
+                objectPath: "/org/freedesktop/portal/desktop",
+                interfaceName: "org.freedesktop.portal.Settings",
+                methodName: "Read",
+                parameters: parameters,
+                replyType: VariantType.New("(v)"),
+                flags: DBusCallFlags.None,
+                timeoutMsec: 2000,
+                cancellable: null
+            );
+
+            uint userThemePreference = ret.GetChildValue(0).GetVariant().GetVariant().GetUint32();
+
+            return userThemePreference switch
+            {
+                1 => AvailableApplicationTheme.Dark,
+                2 => AvailableApplicationTheme.Light,
+                _ => _gtkSettings.GtkApplicationPreferDarkTheme ||
+                     (_gtkSettings.GtkThemeName?.Contains("Dark", StringComparison.OrdinalIgnoreCase) ?? false)
+                    ? AvailableApplicationTheme.Dark
+                    : AvailableApplicationTheme.Light
+            };
+        }
+        catch (Exception ex)
+        {
+            LogGetLinuxThemeFailed(ex);
+            return AvailableApplicationTheme.Light;
+        }
     }
+
+    [LoggerMessage(0, LogLevel.Error, "Failed to detect Linux theme.")]
+    partial void LogGetLinuxThemeFailed(Exception ex);
 }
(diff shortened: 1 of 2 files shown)


CodedOre:
Looks good to me. On Fedora, the app now follows dark mode again, but the titlebar remains light:

Bildschirmfoto vom 2024-08-17 06-01-01

Which I guess is to expected since GTK itself still doesn't know about the dark mode preference...

src/app/dev/platforms/desktop/DevToys.Linux/LinuxProgram.cs
```diff
@@ -31,6 +31,7 @@ internal partial class LinuxProgram
 
     internal LinuxProgram()
     {
+        Gio.Module.Initialize();
```
badcel, line 34: You can remove this line. This is part of the sample as the Gio.Module is not initialized automatically. 

For Libadwaita and GTK applications this is done automatically via the `Gtk.Applocation` class for all GTK dependencies including Gio.

- [Gtk.Application](https://github.com/gircore/gir.core/blob/main/src/Libs/Gtk-4.0/Public/Application.cs) calls `Gtk.Module.Initialize`
- `Gtk.Module.Initialize` calls [Gdk.Module.Initialize](https://github.com/gircore/gir.core/blob/295288990b0f2972d0cbf898c937eea48f6109b8/src/Libs/Gtk-4.0/Public/Module.cs#L35)
- `Gdk.Module.Initialze` calls [Gio.Module.Initialize](https://github.com/gircore/gir.core/blob/295288990b0f2972d0cbf898c937eea48f6109b8/src/Libs/Gdk-4.0/Public/Module.cs#L38)

badcel (commented):
This looks good at a first glance. You should be able to remove the Gio initialization though.

I will try to test your change on my hardware and report back.

badcel:
@CodedOre Thanks for testing then I will skip the testing on my side.

Making the title bar dark should be possible, too.

CodedOre:
Looking through the docs, I think setting the [`gtk-application-prefer-dark-theme` property](https://docs.gtk.org/gtk4/property.Settings.gtk-application-prefer-dark-theme.html) should work?

Though I haven't tried that yet.

badcel:
@CodedOre Yes this works, tested on fedora 40 with:

```csharp
var application = Gtk.Application.New("org.gir.core", Gio.ApplicationFlags.FlagsNone);
application.OnActivate += (sender, args) =>
{
    var window = Gtk.ApplicationWindow.New((Gtk.Application) sender);
    window.Title = "Gtk4 Window";
    window.SetDefaultSize(300, 300);
    window.GetSettings().GtkApplicationPreferDarkTheme = true;
    window.Show();
};

return application.RunWithSynchronizationContext(null);
```

So setting `GtkApplicationPreferDarkTheme ` to true looks to be the right thing to do.

badcel:
I think `Gtk.Settings.GtkApplicationPreferDarkTheme` should not be used as a fallback decider for the preferred theming mode. This is implied by the linked docs of @CodedOre, too.

I think it is more meant as an "application level setting" defining **the developers** preference of either a dark or light theme. E.g. in GNOME there are apps, which use a dark theme, even if the user chooses to have a light theme. This is the case for media related apps like video players. I think those apps set this setting to indicate that **they** prefer a dark theme over the user choise.

And If your code detects, the user wants to use a dark theme you set this property to  true to adhere to the users wish resulting in a dark titlebar.

badcel:
@veler Are you aware that the dbus call might fail regularly in case of old distributions (API not available). In those cases you would not check the GTK theme currently.

veler:
Hey everyone! Thank you all for your feedback, it's super helpful :)
I think I updated the PR according to your suggestions. I'm struggling at setting up a Fedora somehow. I wouldn't mind if one of you could try it. On my side, I tested on Mint and Ubuntu.

veler:
Alright, I finally got Fedora to work!
The title bar now follows the theme of DevToys. However, DevToys does not detect the system's theme. `userThemePreference` on line 179 is `0`, so it fallback on checking the theme name, which is just "adwaita", whether it's dark or not.

I'm happy to take suggestions, otherwise I kind of happy with the current state too. I know it's not perfect but it's not a blocking issue.

CodedOre:
Which edition of Fedora did you install?

Because the Settings portal, like the other XDG desktop portals, depend on a implementation in the desktop to provide the information. 

Gnome and Elementary have implemented the new portal, as well as KDE (partially at least).
But some like Cinnamon or Xfce might not provide the preference yet, so there a fallback to the theme name would be used.

CodedOre:
I did just check with my Fedora desktop (to be precise, Fedora Silverblue with the Gnome desktop), and it does work fine and is indeed using the portal.

badcel:
@veler you could double check if you use "D-Spy" to execute the call to the portal interface manually like [I showed in the screenshot](https://github.com/DevToys-app/DevToys/issues/1381#issuecomment-2294336617).

Perhaps you really have some variant / version which does not have this feature available?

veler:
Hey there,
Thanks for your comments :)

I made a virtual machine of `Fedora Workstation 40` using the ISO from the Fedora website.
If I'm not wrong, @badcel , I'm getting the same result than you in D-Bus. This is in my Fedora, currently set to Dark mode.

image

Am I understanding correctly it finds "1" ? Also, when I switch the theme to Default, then the value is "0".

image

CodedOre:
The values from the settings portal looks right to me. If you look at the [portal documentation](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Settings.html), the value 1 stands for dark mode, while 0 for "no preference". The reason Gnome sets "no preference" instead of "prefer light" on the default setting is that some apps on default would prefer to be dark (like image viewers), while most apps use light as default.

CodedOre:
So, I so notice that as of now, on the default setting the app uses the fallback. So, this could be improved by calling the fallback if the DBus call returns an error (like when it did not find the portal).

src/app/dev/platforms/desktop/DevToys.Linux/Core/ThemeListener.cs
```diff
@@ -147,8 +155,48 @@ private void UpdateSystemSettingsAndApplyTheme()
...
+                interfaceName: "org.freedesktop.portal.Settings",
+                methodName: "Read",
+                parameters: parameters,
+                replyType: VariantType.New("(v)"),
+                flags: DBusCallFlags.None,
+                timeoutMsec: 2000,
+                cancellable: null
+            );
+
+            uint userThemePreference = ret.GetChildValue(0).GetVariant().GetVariant().GetUint32();
```
CodedOre, line 177: Somewhere here you should, if I read the documentation right, check for an G_IO_ERROR_CLOSED error, in which case the fallback should be called.

veler: Thanks ! (and sorry for the late reply) 
I will try to do this soon. :)

badcel: If an error occurs, the `bus.CallSync()` call will throw an exception as gir.core interpretes the error argument by itself and raises an exception. As there is already a `try/catch` in place this should be fine.

So you probably only need to explicitly define in the switch expression which Theme DevToys should use if the user has no preference instead of calling the fallback.

veler: Hello,
So sorry for taking so long to get back to you. I couldn't work on DevToys for about 2 months as I was very busy with other things in my life.
I applied the last recommendation you made. I think we're good to go here :)

